PR #9, opened 2019-05-14
Make menu icon (admin bar & main admin menu) use the menu icon color that's set in the site admin styling (also implement SVG instead of PNG in the process to support this)

KZeni:
The icon needs to be called inline so WordPress then swaps out the SVG fill color for whatever menu icon color it should be (see https://stackoverflow.com/a/42265057/1275054 for more details).

Also, I added menuicon.svg to the assets folder for reference (though it isn't actually called directly since the icon color is determined by having the SVG inline in the different menu item parameters.) Figured it might be helpful to have that icon available as an SVG file (though having menuicon.svg nor w3tc-sprite.png & w3tc-sprite-retina.png, for that matter, isn't required and I could just as easily see these removed from the plugin).

Regarding the icon itself, I tried to replicate the current menu icon while giving it a nice presence. I tried out a few different setups and found that the 3 sides to the box having opacity serve as shading with the shadow then very low opacity for effect (there wasn't a way to get the shadow to stay dark with svg-painter.js colorizing things.)


maxicus:
Thanks.
As for now it contains bug - icon goes black when any frontend page is opened as admin:

w3tc-icon

foreground color is replaced well for backend /wp-admin/* pages only.

maxicus:
Adjusted it a bit to using font, so that it listens to wordpress admin-bar's css.

KZeni:
Yeah, it appears the SVG conversion WordPress does for the primary admin menu to have it use the correct colorization, unfortunately, doesn't extend to the admin bar on the non-admin/dashboard pages of the site. After poking around, it does appear that converting it to an icon font & serving it up that way seems to be the ideal solution for now (having it so both menus then use the font implementation also making sense).

I was a bit bummed by the fact that the resulting icon font didn't have variable opacity for the different elements like the SVG graphic has. As such, I've implemented the icon using https://icomoon.io/app/#/select/font since it has multi-color support during SVG conversion which would retain the opacity of the different elements. I'm currently working on getting it tested and ready to provide.

PR #11, opened 2019-05-14
Improve the rendering of the icon used in the admin menu & admin bar

KZeni:
As mentioned via https://github.com/W3EDGE/w3-total-cache/pull/9#issuecomment-492330248, I've investigated a way to have it keep the opacity values for different parts of the icon and also retain the dynamic coloring of the icon per the icon color that's set for both the admin bar & admin menu (on site admin & front-end pages). I even got it to keep the shadow as black if that's desired in the full icon implementation.

First, I converted the original SVG into an icon font using [IcoMoon.io](https://icomoon.io/app/#/select/font), and this supports multiple glyphs in icons so that was then used moving forward for all of this. The export of the IcoMoon font is available here: [w3tc.zip](https://github.com/W3EDGE/w3-total-cache/files/3179232/w3tc.zip) Additional info/etc. can be provided as needed.

I then implemented the multi-glyph icon for the admin bar to be the ideal setup (it even keeps the shadow as black while the rest changes based on the icon color). However, the admin menu doesn't have enough selectors to use the glyph-based setup so I implemented that using the previous inline SVG setup since that did work fine & it has better handling of the varying opacity on elements. It might be good to revisit this to make it so the inline SVG is replaced with the glyph-based icon when it's possible (let me know if I actually overlooked how it would be now).


maxicus:
We will keep things similar to WordPress core, and since font icon is used there. That way it's better compatible with future WordPress updates.

Looks like shadows can't be implemented in this mode without all those hacks, so we will leave icon without shadows.

If you want to change icon so it looks better with actual code (i.e. make it nicer without shadow) - you are welcome.

PR #15, opened 2019-06-05
set empty strings to empty arrays - fixes #14

### LINKED ISSUE #14: Support page produces PHP warnings
Steps to reproduce:
1. Install and activate W3TC.
2. Go to _Performance > Support_ page in WP dashboard.

Tested in PHP 7.1, and PHP 7.2 - received 4 warnings on each.

Can [use this link to reproduce](https://boldgrid.com/central/cloud-wordpress/try?theme=https:%2F%2Fwordpress.org%2Fthemes%2Fdownload%2Fcrio.2.2.1.zip&plugins=query-monitor,w3-total-cache) in test environment.

PHP warnings:
image

Warning | Cannot assign an empty string to a string offset | 4 | -do_action('admin_print_scripts-performance_page_w3tc_support')wp-admin/admin-header.php:125W3TC\Support_Page::admin_print_scripts_w3tc_support()wp-includes/class-wp-hook.php:286wp_localize_script()wp-content/plugins/w3-total-cache/Support_Page.php:27WP_Scripts->localize()wp-includes/functions.wp-scripts.php:200wp-includes/class.wp-scripts.php:454 | Plugin: w3-total-cache
-- | -- | -- | -- | --

timelsass:
I think this is PHP 7+ issue only, but all the uses of these variables here look like they pass through `encodeURI( $var )` in JS so an empty array will evaluate to an empty string in that case producing the same results.

### DIFF
--- a/Support_Page.php
+++ b/Support_Page.php
@@ -22,11 +22,14 @@ class Support_Page {
 			'w3tc_support_home_url', $url );
 		wp_localize_script( 'w3tc-options',
 			'w3tc_support_email', get_bloginfo( 'admin_email' ) );
+		
 		$u = wp_get_current_user();
-		wp_localize_script( 'w3tc-options',
-			'w3tc_support_first_name', $u->first_name );
-		wp_localize_script( 'w3tc-options',
-			'w3tc_support_last_name', $u->last_name );
+
+		$u_first_name = empty( $u->first_name ) ? array() : $u->first_name;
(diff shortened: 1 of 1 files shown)


timelsass:
https://github.com/W3EDGE/w3-total-cache/commit/a2c24a406b472d1d91f78e6aa6f4ef016f9e53fd is the right way to handle it, wasn't sure if there was reason for being individual vars, closing this PR, thanks!

PR #18, opened 2019-06-15
Enable caching for webp MIME type

njlynch:
Currently, webp is included in the "other_compression" MIME type list, which
results in the ETag and Last-Modified headers being unset in the generated
.htaccess Browser Cache configuration.

This change removes webp from the "other_compression" mime type and enables
webp images to be probably cached by the browser.


maxicus:
Right. Thanks.

PR #20, opened 2019-07-09
Fix the CSS scoping of icon on W3TC site admin pages. Remove shadow from icon font.

KZeni:
This fixes the icon being applied to all `.wp-menu-image` elements on the W3TC site admin pages. The icon has also been updated to not include the shadow since it doesn't quite look right without translucency.

In short, this fixes a CSS issue and makes the icon itself more presentable.


maxicus:
Thanks for participation!

PR #37, opened 2019-09-10
Minify Debug Mode: CSS line numbers debug data problem when source contatins comment with url

SultanICQ:
If you have a block comment in one line with an url inside it, the code incorrectly thinks the comment is not finished.

This is due to the double slash in http:// that gets trim if at the same line there is an */ following it.

For example, this line:

/* Generated by Font Squirrel (http://www.fontsquirrel.com) on May 30, 2013 */

Gets converted to:

/* Generated by Font Squirrel (http:

So it thinks the comment is not finished.

That behaviour breaks web rendering because large portions of valid CSS get deactivated inside a comment when combining them inside the big css with line comments.

### DIFF
--- a/lib/Minify/Minify/Lines.php
+++ b/lib/Minify/Minify/Lines.php
@@ -95,7 +95,7 @@ class Minify_Lines {
     private static function _eolInComment($line, $inComment)
     {
         // crude way to avoid things like // */
-        $line = preg_replace('~//.*?(\\*/|/\\*).*~', '', $line);
+        $line = preg_replace('~(?<!:)//.*?(\\*/|/\\*).*~', '', $line);
 
         while (strlen($line)) {
             $search = $inComment


bwmarkle:
Hi @SultanICQ,

Thanks for the PR! I just wanted to touch base and let you know we'll be taking a look at your pull request shortly, and should have an update for you soon.

Thanks,
- Brad

SultanICQ:
Hello! It works perfectly.

Looking at the code I see that already take into account this case within others.

Regards,

bwmarkle:
Much appreciated @SultanICQ for taking the time to help us test! We're going to wrap up our own testing on this, and hopefully we'll be able to get this fix released in our next patch release.

maxicus:
Thank you for you help!
Applied that by https://github.com/W3EDGE/w3-total-cache/commit/c0379246735bf9e2bd2057d456037ed548872249

Issue #27, opened 2019-08-12
S3 metadata


jpb10:
Since June 2019, files uploaded to S3 have different metadata than before.

Older files:
```
Cache-Control 
Content-Type 
Expires
```

Newer files:
```
Content-Type	
x-amz-meta-access-control-allow-origin
x-amz-meta-content-type
x-amz-meta-last-modified
x-amz-meta-link
```

Which is problematic, because there are no expires or cache control values (for browser caching).

Also, the x-amz-meta-* values seem either redundant or unnecessary.

I suppose changes were made:
https://github.com/W3EDGE/w3-total-cache/commit/49dfa53324d1884796a910d27bf1647abd8227bf#diff-7971eb11690e9cb546b92abb2f8e6d63

maxicus:
Cache-Control header is successfully returned by S3 at the time of issue creation. Code doing it:
https://github.com/W3EDGE/w3-total-cache/blob/master/CdnEngine_S3.php#L263

"Expires" header should not be copied to S3 since it's static afterwards. Max-age should be used (i.e. Cache-Control header).

Remaining handled by:
https://github.com/W3EDGE/w3-total-cache/commit/f9ef3871e5e66bd406a8675cd2adfa5b8fd593a9

x-amz are really useless. Removed.

CORS header can't be set by actual S3 API anymore, instead CORS policy should be set at bucket level. That's done now by "Create bucket" button.

maxicus:
@seang96 
So now you should check why it's empty. At first make sure browsercache.enabled is true, that this is called:
```
			if ( $this->_config->get_boolean( 'browsercache.enabled' ) ) {
				$w3_plugin_browsercache = Dispatcher::component( 'BrowserCache_Plugin' );
				$cdn[0]->cache_config = $w3_plugin_browsercache->get_cache_config();
			}
```

and if so - go to `get_cache_config`. The only thing it does is reads configuration, so I guess you'll have some config options not set. checking it right in the code will show you which exactly.

seang96:
@maxicus I found the issue.

Cdn_Core.php only sets the cache_config for $cdn[0], which in this case is set to CdnEngine_CloudFront.

https://github.com/W3EDGE/w3-total-cache/blob/ec453ed8758e705ce2475915398bc2419b13de09/Cdn_Core.php#L621-L624

[...]

PR #47, opened 2019-09-18
Fix Cache Policy with CDN CloudFront and S3

seang96:
This fixes issue #27 by setting CdnEngine_CloudFront to be a subclass of CdnEngine_S3 which in effect allows cache_config to be set for the entire object and reduces the code needed to be covered in CdnEngine_CloudFront.


bwmarkle:
Hi @seang96 thanks for the pull request. I'm ecstatic to see users like yourself beginning to submit them to the public repo! I'm touching base with the developers on this one, and hopefully soon we'll have some feedback for you.

Thanks!
\- Brad

maxicus:
Fixed differently, anyway thank you for your detailed report about problem.

PR #54, opened 2019-09-25
add config to make memcached binary protocol optional

romainmenke:
w3-total-cache fails to connect to memcached servers with binary protocol turned on.

The cause of this issue is the `\Memcached::OPT_BINARY_PROTOCOL` constant which is defined if there is support for this feature on the server running the Wordpress instance. It is however possible that the actual memcached server instance, or a proxy in between does not support it.

This change adds a config option to disable the binary protocol for memcached.

I'm not a 100% sure I updated all the relevant parts to support this new config. The core functionality is tested and works for us. This setting appears in the Admin interface and is editable there. Feedback is welcome!

### DIFF
--- a/Cache_Memcached.php
+++ b/Cache_Memcached.php
@@ -59,8 +59,12 @@ class Cache_Memcached extends Cache_Base {
 		if ( defined( '\Memcached::OPT_REMOVE_FAILED_SERVERS' ) ) {
 			$this->_memcache->setOption( \Memcached::OPT_REMOVE_FAILED_SERVERS, true );
 		}
-		if ( defined( '\Memcached::OPT_BINARY_PROTOCOL' ) && defined( '\Memcached::OPT_TCP_NODELAY' ) ) {
+
+		if ( isset( $config['binary_protocol'] ) && !empty( $config['binary_protocol'] ) && defined( '\Memcached::OPT_BINARY_PROTOCOL' ) ) {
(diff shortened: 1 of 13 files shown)


maxicus:
Thanks. We wanted to avoid one more config option (too much already) but seems there are still a lot legacy systems...

How about moving OPT_TCP_NODELAY there too? In that case disabling binary protocol would switch all those new features.
Does it work well for you without OPT_TCP_NODELAY?

romainmenke:
As far as we can tell it works just fine with or without `OPT_TCP_NODELAY` (we did not run our benchmarks). It only breaks with `OPT_BINARY_PROTOCOL` on.

Maybe keep these features separate for now? We understand wanting to keep the config options to a minimum but throwing in `OPT_TCP_NODELAY ` behind a config options `binary_protocol` seems confusing and counter intuitive to us.

romainmenke:
@maxicus What is your preferred fix for this issue? I don't mind putting more work into this PR to make sure you are happy with it.

romainmenke:
@maxicus ping

maxicus:
@romainmenke Thank you for your help. 
We just giving a chance for other people to report something related to this area.
The primary question if anyone has problems with NO_DELAY mode.

So far your fix is fine, and we are going to merge it to master shortly. In any case before the next release.

Issue #52, opened 2019-09-24
FAQ Link in Dashboard should open in new tab


harryjackson1221:
In the Dashboard, the FAQ link opens in the same tab, causing you to be taken away from your website when navigating to it.

![image](https://user-images.githubusercontent.com/29436971/65516875-c529ba00-deaf-11e9-9d1a-e9080317ca78.png)

Adding a target="_blank" to the link would allow the user to see the FAQ, and not navigate away from their site in the process.

PR #64, opened 2019-10-09
FAQ Link in Dashboard open in new tab

raphael-concil:
Change FAQ link in left sidebar to open in a new tab via JQuery.
Change FAQ link on top admin bar to open in new tab with target attribute.
This commit fix issue #52


maxicus:
Thanks for idea, looks like it's that hack is the only way to manipulate with menu.
Applied it with some clean up via https://github.com/W3EDGE/w3-total-cache/commit/18211da75dccaf01fa6d036a3f3325e7718741e2

Issue #92, opened 2019-12-09
PHP 7.4 PHP Deprecated: Function get_magic_quotes_gpc() is deprecated


mavas84:
Reference topic: https://wordpress.org/support/topic/php-deprecated-function-get_magic_quotes_gpc-is-deprecated/

viktorgino:
From the documentation : 

> Returns 0 if magic_quotes_gpc is off, 1 otherwise. Or always returns FALSE as of PHP 5.4.0. 

Source: https://www.php.net/manual/en/function.get-magic-quotes-gpc.php

So the get_magic_quotes_gpc() can be removed altogether. Quick search show that it can be found here:
https://github.com/W3EDGE/w3-total-cache/blob/master/Util_Request.php#L25
https://github.com/W3EDGE/w3-total-cache/blob/master/lib/Minify/HTTP/ConditionalGet.php#L367
https://github.com/W3EDGE/w3-total-cache/blob/master/PgCache_ContentGrabber.php#L1742
https://github.com/W3EDGE/w3-total-cache/blob/master/lib/Db/mssql.php#L522

bwmarkle:
Just a heads up, issues https://github.com/W3EDGE/w3-total-cache/issues/92 and https://github.com/W3EDGE/w3-total-cache/pull/95 are being discussed internally as they cause problems with PHP 5.3, and a decision needs to be made whether we still need to support php 5.3

PR #95, opened 2019-12-12
Remove deprecated get_magic_quotes_gpc()

viktorgino:
fixes https://github.com/W3EDGE/w3-total-cache/issues/92

### DIFF
--- a/PgCache_ContentGrabber.php
+++ b/PgCache_ContentGrabber.php
@@ -1739,7 +1739,7 @@ class PgCache_ContentGrabber {
 	 */
 	function _check_match( $etag ) {
 		if ( !empty( $_SERVER['HTTP_IF_NONE_MATCH'] ) ) {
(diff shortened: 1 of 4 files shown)


bwmarkle:
Just a heads up, issues https://github.com/W3EDGE/w3-total-cache/issues/92 and https://github.com/W3EDGE/w3-total-cache/pull/95 are being discussed internally as they cause problems with PHP 5.3, and a decision needs to be made whether we still need to support php 5.3

maxicus:
That works on most cases in php5.3 too. HTTP_IF_NONE_MATCH check is acceptable to fail at PHP level due to multiplication of probabilities that it's used, it's used in PHP, PHP 5.3 is used and magic quotes enabled and lower risk since doesnt cause critical failures.

Util_Request is fine since TEMPLATEPATH is almost always defined.

Thanks @viktorgino for participation.

Issue #181, opened 2020-04-30
Settings not translated to profil languages


sebastienserre:
Hello,

Since several WP release, user can set a different language than default website language in their profile.

Here my website language is defined to English:
![WP Lang](https://i.imgur.com/VKQ75Od.png)

And my profile language is French
![Profile Lang](https://i.imgur.com/ixfOAfm.png)

So all the back end is  in French, except W3TC menu (and this is the mistake):
![W3TC menu](https://i.imgur.com/2C9wwcQ.png)

W3TC is fully translated to French: https://translate.wordpress.org/locale/fr/default/wp-plugins/w3-total-cache/

bwmarkle:
Hi @sebastienserre thanks for reporting this issue! Just wanted to let you know we've taken a look at your ticket, and we'll get back to you soon.

sebastienserre:
Hello,
I dig in the code and found some problem.
Some details are given in this link to improve the i18n:
https://wp-info.org/tools/checkplugini18n.php?slug=w3-total-cache
On this screenshot:
![2020-05-08_16-39](https://user-images.githubusercontent.com/3902621/81416851-9dd7e280-914a-11ea-8ddf-05cd39b329e0.png)
You can see a mix between english and French.

The untranslated sentence "Reduce load time by decreasing the size and number of CSS and JS files. Automatically remove unncessary data from CSS, JS, feed, page and post HTML." is powered by your own function `w3tc_e()` here:
https://github.com/W3EDGE/w3-total-cache/blob/0013107b3d1c8fffe2c570ffec754960849832cf/inc/options/general.php#L129

If we change this line to
`		<p><?php w3tc_er( 'minify.general.header', _e('Reduce load time by decreasing the size and number of <acronym title="Cascading Style Sheet">CSS</acronym> and <acronym title="JavaScript">JS</acronym> files. Automatically remove unncessary data from <acronym title="Cascading Style Sheet">CSS</acronym>, <acronym title="JavaScript">JS</acronym>, feed, page and post <acronym title="Hypertext Markup Language">HTML</acronym>.', 'w3-total-cache' ) ) ?></p>
`
Then it's translated:
![2020-05-08_17-02](https://user-images.githubusercontent.com/3902621/81418933-bb5a7b80-914d-11ea-9934-918b102dd288.png)

cssjoe:
[...]

PR #185, opened 2020-05-08
i18n Improvment

sebastienserre:
Related to #181


maxicus:
Hi,

Thanks for pull request. In general it's fine.
Just can you remove formatting-only changes from diff?
In `SystemOpCache_GeneralPage_View.php` you seems changed only 1 line of code, while there are much more formatting-only changes not related to the topic of this pull request directly.

It makes it difficult for people approving it and affects blame without a reason.

Issue #146, opened 2020-02-26
Add Filter supporting Shortpixel to replace the image URL and then send back the adjusted CSS


mavas84:
Add filter inside the plugin code for the following scenario:
- a user who is using W3TC plugin and also uses ShortPixel Adaptive Images plugin
- the site of this user is also using the CSS combine & minify options from W3TC plugin
- inside the CSS of the site, there are various calls to background images which are processed by ShortPixel Adaptive Images

In this scenario, Shortpixel will extract the CSS containing images and will serve it separately, making some additional requests which could be avoided. In order to avoid this situation which is not optimal for the user (from a performance point of view),  a filter is needed in your W3TC, which would allow Shortpixel to replace the image URL and then send back the adjusted CSS so that W3TC can do its job with the combine and minify modules. The filter needs to have the following:
- the CSS that will be combined/minified
- the source path of the original CSS, before combining/minifying (if available)
- the target path (on disk) where the combined CSS is written on disk

pdobrescu:
thanks for adding the issue @mavas84! In case you need further details, please leave a comment here.

pdobrescu:
We have just submitted pull request https://github.com/W3EDGE/w3-total-cache/pull/191 for this issue, please check it and review it. Thanks!

cssjoe:
Thank you @pdobrescu and @short-pixel-optimizer for the contribution, which was released in version 0.14.0 on May 27th.

PR #191, opened 2020-05-20
Apply a filter on the minified content - w3tc_minify_css_content

short-pixel-optimizer:
This enables  image optimizers such as ShortPixel Adaptive Images to dynamically replace the background-image URLs in the W3 Total Cache cached versions of the CSS file, eliminating the need for another level of caching.

Fixes https://github.com/W3EDGE/w3-total-cache/issues/146


short-pixel-optimizer:
Thanks @cssjoe for the feedback! Made the change.

cssjoe:
Thank you @pdobrescu and @short-pixel-optimizer for the contribution, which was released in version 0.14.0 on May 27th.

PR #198, opened 2020-05-28
Fix: CSS URL rewrite logic

cssjoe:


marvec:
The fix works. Thank you for a super fast action!

PR #211, opened 2020-06-23
Issue 206

bwmarkle:


inc/lightbox/upgrade.php
```diff
@@ -3,28 +3,37 @@
...
+
+		<div>
+			<div>
+				<strong>W3 TOTAL CACHE</strong><br />
+				<span style="font-size:16px;"><?php esc_html_e( 'Unlock more performance options', 'w3-total-cache' ); ?></span>
+			</div>
+		</div>
+
+		<div>
+			<?php if ( true ): ?>
```
maxicus, line 19: button doesnt work on http:// websites, that's why there was a condition there

bwmarkle: Very nice catch @maxicus, that was my debug code in there... I fixed it in https://github.com/W3EDGE/w3-total-cache/pull/211/commits/c01bf9db1f8aadc7e5a27d69003f2a131f12d39c

PR #230, opened 2020-07-28
Fix compatibility check for Apache servers when php running as CGI

8ctopus:
For Apache servers when php is running as CGI, self-test doesn't detect installed Apache modules.

The objective of this pull request is to detect Apache modules for any php configuration (requires php exec function to be enabled on server)


bwmarkle:
Hi @8ctopus thanks so much for the PR! Sorry for taking so long to get back to you on this... I'll pass this along to @maxicus and @cssjoe to see if they can provide any input.

Thanks!
- Brad

8ctopus:
no problem guys, take your time

PR #240, opened 2020-08-11
Add option to disable jquery-migrate + WP 5.5 updates

cssjoe:


ConfigKeys.php
```diff
@@ -2417,7 +2417,10 @@
 		'default' => '',
 		'master_only' => true
 	),
-
+	'jquerymigrate.disabled' => array(
```
bwmarkle, line 2420: why

Generic_Plugin.php
```diff
@@ -84,6 +84,31 @@ function run() {
 					'ob_callback'
 				) );
 		}
+
+		// If WP < 5.5, disable jquery-migrate on the front-end, if configured.
```
bwmarkle, line 88: why

Generic_Plugin.php
```diff
@@ -84,6 +84,31 @@ function run() {
...
+		if ( ! is_admin() && version_compare( get_bloginfo( 'version' ), '5.5', '<' ) && $this->_config->get_boolean( 'jquerymigrate.disabled' ) ) {
+			add_action( 'wp_default_scripts', array( $this, 'disable_jquery_migrate' ) );
+		}
+	}
+
+	/**
+	 * Disable jquery-migrate.
```
bwmarkle, line 95: why

inc/options/general.php
```diff
@@ -481,6 +481,17 @@ class="w3tc-ignore-change" type="text"
 				</th>
 			</tr>
 			<?php
+			if ( version_compare( get_bloginfo( 'version' ), '5.5', '<' ) ) {
```
bwmarkle, line 484: So we're adding a new setting for users to disable jquery migrate on the front end? I wasn't really aware what was in this ticket.

Generic_Plugin.php
cssjoe: I removed the version check.

Generic_Plugin.php
cssjoe: It was a feature request that was marked Next.

inc/options/general.php
cssjoe: Yes, see the link in this PR.  I removed the version check.  https://make.wordpress.org/core/2020/06/29/updating-jquery-version-shipped-with-wordpress/

maxicus:
The change is fine by idea, just the logic itself shouldn't be in general module, but user experience one.
In general nothing should be in general module :)

Extension_CloudFlare_Page_View.js
```diff
@@ -33,7 +33,7 @@ jQuery(function($) {
 
 
 		.on('click', '.w3tc_cloudflare_zone_page', function() {
-			var page = jQuery(this).attr('data-page');
+			var page = jQuery(this).prop('data-page');
```
maxicus, line 36: That and all similar looks like a bug.
prop shouldnt return value for non-DOM properties.

Did you check it works?

Generic_Plugin.php
```diff
@@ -84,6 +84,31 @@ function run() {
 					'ob_callback'
 				) );
 		}
+
```
maxicus, line 87: there's still code in general.

The idea is that if you pollute general module - everything will be there at some point and it becomes unmaintantable.

Extension_CloudFlare_Page_View.js
cssjoe: I'm fixing all of those "data-" references to see "data()", which works.

Generic_Plugin.php
cssjoe: I moved the code to a new "UserExperience_Plugin" module.

Extension_CloudFlare_View_Dashboard.js
```diff
@@ -1,6 +1,6 @@
 jQuery(function() {
     jQuery('#w3tc_dashboard [type=submit]').bind('click', function(){
-        jQuery(this).attr('was_clicked','yes');
+        jQuery(this).prop('was_clicked','yes');
```
maxicus, line 3: @cssjoe how about that? i guess there are like 100 similar non-default props.

cssjoe: I reverted those.

Root_Loader.php
```diff
@@ -20,6 +20,7 @@ function __construct() {
 
 		$plugins = array();
 		$plugins[] = new Generic_Plugin();
+		$plugins[] = new UserExperience_Plugin();
```
maxicus, line 23: this should be conditional to not load files (and waste compilation+execution time and memory) when they are not used.

see e.g. https://github.com/W3EDGE/w3-total-cache/blob/583d7a056f0d19345ac8045ec175c42b4ba414d1/Root_Loader.php#L35

cssjoe: I renamed the new class and conditionally loaded it.

bwmarkle:
@cssjoe I'm poking around... The first 2 themes I tested aren't loading jQuery migrate on the front end. Can you simply install any theme and you'll see jquery migrate? How exactly was this tested?

UserExperience_Plugin_Jquery.php
```diff
@@ -0,0 +1,67 @@
...
+	 */
+	public function run() {
+		// Disable jquery-migrate on the front-end, if configured.
+		if ( ! is_admin() && $this->_config->get_boolean( 'jquerymigrate.disabled' ) ) {
+			add_action( 'wp_default_scripts', array( $this, 'disable_jquery_migrate' ) );
+		}
+	}
+
+	/**
+	 * Disable jquery-migrate.
```
bwmarkle, line 49: jQuery migrate is not removed if I enqueue it by itself.

```
add_action(
    'wp_enqueue_scripts',
    function() {
	    wp_enqueue_script( 'jquery-migrate' );
    },
    100
);
```

Current approach only works if it's registered as a dependency.

bwmarkle:
When I add this to the functions.php file of twentyeleven

```php
add_action( 'wp_enqueue_scripts', function() {
    wp_register_script( 'jquery-migrate', '//code.jquery.com/jquery-migrate-1.2.1.min.js', array( 'jquery' ), '1.2.1', true );
    wp_enqueue_script( 'jquery-migrate' );
    wp_enqueue_script( 'jquery' );
} );
```

Then do this:
image

I still see this:
image

bwmarkle:
@cssjoe Any thoughts on these errors?

image

cssjoe:
Check again.  The updated file https://github.com/W3EDGE/w3-total-cache/blob/848ec8faea67e4c2ee73f97dcaaba53630663edc/pub/js/options.js here has `.size()` replaced with `.length`.

PR #269, opened 2020-09-17
Update ConfigKeys.php

harryjackson1221:
Added mfunc to default for HTML Minify - Ignore comment stems


harryjackson1221:
https://wordpress.org/support/topic/mfunc-needs-to-be-excluded-from-minifying/

WooCommerce has an article for this also:
https://docs.woocommerce.com/document/configuring-caching-plugins/#section-2

image

cssjoe (approved):

PR #277, opened 2020-09-30
Remove rewrites if disabled. Fixes #275

### LINKED ISSUE #275: Page Cache rewrites persists when disabled
When `Disk: Enhanced` is selected and the settings are saved, mod_rewrite code is added to `.htaccess`.  If Page Cache is disabled, then the rewrites are not removed.   The rewrites are removed or added when changing the engine, regardless if the Page Cache feature is enabled or not.
While Page Cache is enabled and the website front-end is visited, cache files are created in `wp-content/cache/page_enhanced/`.  When Page Cache is disabled, the cache files still exist and are used by the rewrites, which can lead to serving old pages if they were updated since the cache files were created.

The rewrites should be removed when Page Cache is disabled.

cssjoe:


maxicus (changes requested):
also rename it to something like is_rules_required

maxicus (approved):

Issue #281, opened 2020-10-05
Remove serialize / unserialize in Cache_Apcu.php


mavas84:
The current version of W3 Total Cache serializes/unserializes data with the APCu cache backend.

This is unnecessary, apc_store() / apc_fetch() accept mixed data as the value, and APC performs serialization automatically.
W3 Total Cache’s serialization prevents usage of binary ( https://github.com/igbinary/igbinary ) with APC. It is a more efficient plug-in serializer with APC for PHP.
It stores data in binary format, which reduces cache memory footprint significantly. Also, de-serialization speed is better than PHP native deserialization.

The serialization calls should be removed in the cache adapter for better performance. 
Reference topic: https://wordpress.org/support/topic/remove-serialize-unserialize-in-cache_apcu-php/

PR #283, opened 2020-10-05
Remove serialize/unserialize from APCu cache.

mongeron:
https://github.com/W3EDGE/w3-total-cache/issues/281

I have tested these changes briefly with a WordPress installation that has WP object caching enabled, and the caching is used in plugin code. I haven't encountered any issues.

However, proper testing should still be performed against the changes, although they are trivial in a way.

### DIFF
--- a/Cache_Apcu.php
+++ b/Cache_Apcu.php
@@ -42,7 +42,7 @@ class Cache_Apcu extends Cache_Base {
 		$var['key_version'] = $this->_get_key_version( $group );
 
 		$storage_key = $this->get_item_key( $key );
-		return apcu_store( $storage_key, serialize( $var ), $expire );
+		return apcu_store( $storage_key, $var, $expire );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


maxicus (approved):

cssjoe (changes requested):
Our automated tests show "cache key ... no entry found" errors.  When working correctly, the HTML pages should contain "using apc" and not indicate cache key errors.

maxicus (changes requested):
QA results show behavior is unstable.
Most tests fail.

Didn't investigate personally since change is not a priority and not in a roadmap.

mongeron:
Tests are most likely written with different expectations and I didn't check them. I will look into fixing them.

The patched version has been running on production servers since I created the patch so it looks stable to me.

mongeron:
Tests most likely fail because tests are only run using `Cache_Apc` class, while the main plugin is using `Cache_Apcu` class or `Cache_Apc` class, depending on the existence of `apcu_store()` function. This is checked on Cache.php line 32.

I will remove the serialization also from Cache_Apc.php and see if I can run the tests myself.

cssjoe:
@mongeron We will be happy to review your new changes.  Have you had a chance to perform additional testing?  I suspect that most users that enable APC use APCu.

cssjoe:
We can look into this possible optimization in the future if there is a greater need for it, and it can pass all automated tests.

watou:
> Didn't investigate personally why since change is not a priority and not in a roadmap.

Wouldn't ongoing improvements to cache performance and resource usage be on the roadmap?

mongeron:
Sorry for not updating this yet.

I've been busy with other issues, so I haven't managed to check the tests thoroughly. However, I'll take it under work as soon as I get the chance.

I'll open a new PR once I have test coverage fixed.

PR #285, opened 2020-10-06
Add option to minify only for both JS and CSS in auto mode

### LINKED ISSUE #170: JS & CSS Combine
**With current options in auto minification:**
- There is no "minify only" option,
- You have to combine JS in order to defer it,
- To enable the HTTP/2 Push option, you must activate "combine" first.

Manual minify feature is absolutely brilliant and W3TC is the only plugin which offers that option. We can use the "minify only" option if the Minify Mode is manual. However, As theme&plugin assets are frequently changing, people usually tend to use the "Auto minify" mode I think. Like I said we don't have the above features if the minify mode is Auto.

I think, more customization in the minify settings will make the plugin better.

After HTTP/2 improvements, both JS or CSS combine features do NOT have a significant role on page load. Even HTTP/3 is also coming...

Regards.

lwj5:
Hi there,

I've added the option to minify CSS and JS without combining which fixes #170.

However, is there a way to map the values from the current config key to the new config key after a plugin update? 
For example if the current value of `minify.css.combine == true`, I would like to set the intital value of `minify.css.method = 'combine'` else `minify.css.method = 'both'`. And at the same time how can we deprecate or remove `minify.css.combine`.
Can the authors (or other contributors) provide some input on this.

**Benefits**

Fixes #170, benefits described in the issue.

**Future PRs**

- Allow ignoring .min.js and .min.css when minify only.

### DIFF
--- a/ConfigCompiler.php
+++ b/ConfigCompiler.php
@@ -247,6 +247,19 @@ class ConfigCompiler {
 		}
 
 		//
+		// changes in 0.15.2
+		//
+		if ( version_compare( $file_data['version'], '0.15.2', '<' ) ) {
+			if ( isset( $file_data['minify.js.combine.header'] ) && $file_data['minify.js.combine.header'] ) {
+				$file_data['minify.js.method'] = 'combine';
+			}
+
+			if ( isset( $file_data['minify.css.combine'] ) && $file_data['minify.css.combine'] ) {
+				$file_data['minify.css.method'] = 'combine';
+			}
+		}
+
+		//
 		// changes in 0.13
 		//
 		if ( version_compare( $file_data['version'], '0.12.0', '<=' ) ) {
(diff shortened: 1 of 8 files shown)


maxicus:
@lwj5 I like your code, thanks.
Regarding upgrade - you should remove 'minify.css.combine' from ConfigKeys and do the import in 
https://github.com/W3EDGE/w3-total-cache/blob/master/ConfigCompiler.php#L252

import runs when version in your master.php config differs from w3tc version (W3TC_VERSION constant), so you'll need to update your master.php and put some older version to test it.

Let me know if you need a help.

lwj5:
Thanks, @maxicus. I've added the upgrade to ConfigCompiler for version 0.15.2, feel free to change it if the launched version differs.

The PR is complete and ready for review/merging.

maxicus (approved):

PR #286, opened 2020-10-07
Move get_response_headers to Util_Http

ericek111:
There was some redundant code regarding HTTP headers. Now it's more dumb-proof (handles testing environments with XDebug) and generalized.

This commit is a result of a little brain fart, but I figured it'd be beneficial to make a PR anyway.


maxicus:
Thanks for letting know, but growing common code size (memory consumption when pgcache is not enabled) and slowing down runtime by extra debug conditions isn't good.

PR #307, opened 2020-11-25
keep rules consistent with BC cache, fixes #236

### LINKED ISSUE #236: Cache-Control disabled for 304 response when minify is enabled
Browser Cache’ configuration enabled for CSS & JS with:
* Set Last-Modified header: Enabled,
* Set expires header: Enabled,
* Expires header lifetime: 51 – (as for testing purpose),
* Set cache control header: Enabled,
* Cache Control policy: cache with max-age (“public, max-age=EXPIRES_SECONDS”)
* Set entity tag (eTag): Enabled

For the first fresh request for minified CSS file we are getting below ‘Cache-Control’ response from WordPress:
**cache-control: max-age=51, public**
But after expiration of 51 sec. the conditional request is performed, and in the response (304 in case of no change) is always:
**cache-control: max-age=0, public**

When Minify is disabled both 200 and 304 responses are getting **cache-control: max-age=51, public**

Reference topic: https://wordpress.org/support/topic/cache-control-disabled-for-304-response/

maxicus:


cssjoe (approved):
For reference: https://httpd.apache.org/docs/2.4/mod/mod_expires.html#expiresbytype
This changes the `Expires` headers to use the client's access time instead of the file's modification time.

PR #313, opened 2020-12-17
Do not use Redis SETEX when expire=0

BasementCat:
In redis 6 (and possibly earlier), calling SETEX with an expiry of 0 seconds is invalid.  When $expire is 0, call SET instead.


cssjoe (approved):
Thanks, @BasementCat for your contribution.  I made a slight change for WordPress coding standards.

PR #323, opened 2021-01-18
Feature Showcase

cssjoe:


bwmarkle (approved):

Generic_Plugin_Admin.php
```diff
@@ -235,6 +238,20 @@ function admin_enqueue_scripts() {
 		wp_register_script( 'w3tc-widget', plugins_url( 'pub/js/widget.js', W3TC_FILE ), array(), W3TC_VERSION );
 		wp_register_script( 'w3tc-jquery-masonry', plugins_url( 'pub/js/jquery.masonry.min.js', W3TC_FILE ), array( 'jquery' ), W3TC_VERSION );
 
+		// New feature count for the Feature Showcase.
+		wp_register_script( 'w3tc-feature-counter', plugins_url( 'pub/js/feature-counter.js', W3TC_FILE ), array(), W3TC_VERSION, true );
```
maxicus, line 242: @cssjoe its better to avoid pushing module-specific stuff into generic code.
breaks modularity.

Generic_Plugin_Admin.php
```diff
@@ -409,25 +426,34 @@ function admin_print_styles() {
...
+			);
 		}
 
-
 		switch ( $this->_page ) {
 		case 'w3tc_minify':
-		case 'w3tc_mobile':
-		case 'w3tc_referrer':
+		case 'w3tc_cachegroups':
+			wp_enqueue_script(
```
maxicus, line 447: the same here.
js for specific pages are loaded in plugin code itself for later code.

inc/wizard/template.php
```diff
@@ -51,14 +51,8 @@ public function __construct( array $config ) {
 	 * Render the wizard.
 	 *
 	 * @since 2.0.0
-	 *
```
maxicus, line 54: not module-isolated, will discuss

maxicus (changes requested):

BrowserCache_Page.php
```diff
@@ -1,27 +1,20 @@
...
 namespace W3TC;
 
-
-
 class BrowserCache_Page extends Base_Page_Settings {
 	protected $_page = 'w3tc_browsercache';
 
-
-	static public function w3tc_ajax() {
+	public static function w3tc_ajax() {
```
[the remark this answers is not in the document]
cssjoe: That was from PSR2, which doesn't really matter.  I'll skip those in the future.
>The static declaration must come after the visibility declaration
>(PSR2.Methods.MethodDeclaration.StaticBeforeVisibility)

Generic_Plugin_Admin.php
cssjoe: I can move pub/js/feature-counter.js to Generaic_AdminFeatureShowcase.js in a patch release.

PR #324, opened 2021-01-20
Allow cache in PHP-CLI/console and fix bugs when clearing cache from PHP-CLI/console

metiseducation:
1. Allow cache to store new values when in PHP-CLI/console environment.
In our use case we have a time consuming cron-evebt that regularily updates a value in the object cache. We only run cron events from wp-cli, not webrequests.

2. Use wordpress option value 'siteurl' instead of $_SERVER['HTTP_HOST'].
This is important because when running from PHP-CLI/console, the $_SERVER['HTTP_HOST'] value is empty.
Because $_SERVER['HTTP_HOST'] is empty, the cache keys for e.g memcached will be different from the keys from regular webrequests.
This commit will fix problems when attempting to flush cache values from PHP-CLI/console environment.

### DIFF
--- a/CdnEngine_Mirror_CloudFront.php
+++ b/CdnEngine_Mirror_CloudFront.php
@@ -49,7 +49,7 @@ class CdnEngine_Mirror_CloudFront extends CdnEngine_Mirror {
 	 * @return string
 	 */
 	function _get_origin() {
-		return Util_Environment::host_port();
+		return Util_Environment::host();
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


maxicus (changes requested):
1. cache based on actual host by intent since whats requested by user is not always siteurl (common setup).
2. it breaks support for non-default port setups.
3. Util_Environment calls are done before db is available.

royhansen-metis:
@maxicus I understand my approach will not work.
I sent a new pull request where I propose adding a new optional 'host' config option instead: #326

maxicus:
@royhansen-metis yeah, some parameter to wp-cli looks real.
it already has --url parameter btw https://make.wordpress.org/cli/handbook/references/config/

But not sure if it works out of the box for your case or some adjustments needed.

royhansen-metis:
@maxicus I am aware of the --url parameter for wp-cli, however it would be nice to just put a optional 'host' value in the w3tc config.

PR #326, opened 2021-01-20
Allow cache in PHP-CLI/console and add config parameter 'host'

metiseducation:
Allow cache to store new values when in PHP-CLI/console environment.
In our use case we have a time consuming cron-event that regularily updates a value in the object cache. We only run cron events from wp-cli, not webrequests.

I propose to add support for a new config option 'host', which will be optional.
Use w3-total-cache config value 'host' instead of $_SERVER['HTTP_HOST'], if it is present.
This is important because when running from PHP-CLI/console, the $_SERVER['HTTP_HOST'] value is empty, unless url is specified in the wp-cli command.
When $_SERVER['HTTP_HOST'] is empty, the cache keys for e.g memcached will be different from the keys from regular webrequests


maxicus:
@metiseducation Hmm, I tried to find a case which that solution solves, but wp-cli's --url parameter doesn't.
Am I missing something?

royhansen-metis:
With a 'host' value in w3tc config, the value could be automatically set for non-multisite installations, based on the wordpress 'siteurl' option value.
On multisite installations the value would not be set, in these cases one should use the url-parameter.

On a  non-multisite installation it dosn't make sense to type the --url parameter when running wp-cli. That's why this change makes sense.

cssjoe (approved):

cssjoe:
@metiseducation Thank you for your contribution.  It has been a while and we performed a lot of updates since this pull request has been created.  The effective merge will allow WP-CLI calls to cause object cache updates and not be rejected.

metiseducation:
I reviewed the changes you pulled. However I noticed that you did not pull everything from my commit. With your changes, if you run a wp-cli command, the host value will be empty, thus the cache that is read/written in wp-cli will be different from the cache accessed from a browser. Because the browser is using the $_SERVER['HTTP_HOST'] value. So summarize, your changes will allow cache to be used in wp-cli, but the cache read/write will be a separate cache that is only available in wp-cli, and will not be available from the browser.

PR #334, opened 2021-02-05
Namespace W3TC libraries to avoid conflicts with other plugins. Minify & yui library

maxicus:


cssjoe (approved):
I had to fix a missing backslash for csstidy and change HTTP to HTTPS for the JS Closure Compiler.

PR #336, opened 2021-02-09
Add video/ogg support for browser chaging

wittich:
At the moment W3TC doesn't support Browser Cache via .htaccess for video/ogg files. The pull request solves this. 

Therewith the expiring header is also set for those video files.


maxicus:
Thanks!

PR #342, opened 2021-02-19
fix file permissions

romainmenke:
We recently started checking for files that have executable permissions and discovered these.
Is it correct that these are all accidentally executable?

We use a github action to check for this automatically and are happy to share this.


cssjoe:
The file modes may be set if you clone/checkout from GitHub and your config file has `filemode = true` in the `[core]` section, else the files get the default mode depending on your environment's `umask` setting.
Regardless, it is ok for the files to be mode 644 or higher; even 755 (executable).  We have updated several file modes on our master branch to make the modes consistent.  There were more than listed in your PR.
Thank you for looking into possible issues.  Please let us know if there is anything else that we can do to improve W3 Total Cache.

Issue #344, opened 2021-02-25
Stop flushing whole object cache on post update


mavas84:
The object cache is flushed on EVERY post update. 
Add ability to clear the Object cache only on the specific page once the post is updated.
Reference link: https://wordpress.org/support/topic/stop-flushing-whole-object-cache-on-post-update/

archon810:
Flushing the whole object cache on every little post update makes object cache much less useful as it constantly needs to be recached, potentially with expensive and long queries.

maciejmackowiak:
@mavas84 How about using the purge all objects setting which is already in the plugin:
https://github.com/W3EDGE/w3-total-cache/pull/345

archon810:
Any updates please?

archon810:
Any updates please?

Seems like the fix is ready but it keeps getting punted.

PR #345, opened 2021-02-26
Prevent flushing whole object cache on post update

maciejmackowiak:
https://github.com/W3EDGE/w3-total-cache/issues/344

### DIFF
--- a/ObjectCache_Plugin.php
+++ b/ObjectCache_Plugin.php
@@ -338,7 +338,6 @@ class ObjectCache_Plugin {
 	private function _do_flush() {
 		//TODO: Requires admin flush until OC can make changes in Admin backend
 		return $this->_config->get_boolean( 'cluster.messagebus.enabled' )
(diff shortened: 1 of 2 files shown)


maxicus:
The reason why w3tc does it is because many themes store post-related stuff in object cache and don't purge/refresh it on post updates.

i.e. that should be a separate objectcache.purge.wp_admin_change option enabled by default.

maxicus (changes requested):
The reason why w3tc does it is because many themes store post-related stuff in object cache and don't purge/refresh it on post updates.

i.e. that should be a separate objectcache.purge.wp_admin_change option enabled by default.

cssjoe:
@maciejmackowiak Thank you for your contribution.  We have been working on improving the object cache and this pull request has become obsolete.  Please take a look at https://github.com/W3EDGE/w3-total-cache/pull/436 which does handle flush on update.

PR #354, opened 2021-03-26
Fix issue 6

cssjoe:
This fix addresses the issue of an incorrect server document root, broken loading for PHP Tidy, and a CDATA wrapper that caused syntax errors.

To simulate an incorrect server document root path, use the following code in `wp-config.php`:
```
$_SERVER['DOCUMENT_ROOT'] = '/var/www/html';
```

### DIFF
--- a/ConfigKeys.php
+++ b/ConfigKeys.php
@@ -161,6 +161,11 @@ $keys = array(
 		)
 	),
 
+	'docroot_fix.enable' => array(
+		'type' => 'boolean',
+		'default' => false,
+	),
+
 	'lazyload.enabled' => array(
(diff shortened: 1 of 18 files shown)


maxicus:
That new option requires user to know his docroot and know that it's incorrect.
If that's the case - instead of placing correct value to some config he may put it to server config, or, if not available due to hosting plan - do exactly the same as in original reproduction case - change it in wp-config.php :)
Doing that he will also fix all the problems with WordPress itself and other plugins so that's much more preferred method. 

IMHO better to add that to docs instead of adding redundant complexity.

PR #400, opened 2021-07-25
Allow use of default AWS credential provider

### LINKED ISSUE #399: Allow setting AWS IAM permissions through an instance profile
Please support the use of the AWS SDK's default credential provider, rather than requiring that S3 and CloudFront CDN users pass AWS credentials into the plugin's configuration.  I expect the most straightforward way to implement this change is to

1.  allow users to leave the fields for the AWS access and secret key blank (which may or may not require changes to input validation)

2.  in the files which set AWS credentials ([1][1], [2][2], [3][3], [4][4], [5][5], [6][6]), check whether the key and secret are both blank, and if so, instead set `$credentials = \Aws\Credentials\CredentialProvider::defaultProvider();` as described in mkostrikin's comment [here][comment].

[1]: https://github.com/W3EDGE/w3-total-cache/blob/73911133ca843de9c20086c73aa84b134f4f9bb4/CdnEngine_CloudFront.php#L33
[2]: https://github.com/W3EDGE/w3-total-cache/blob/73911133ca843de9c20086c73aa84b134f4f9bb4/CdnEngine_Mirror_CloudFront.php#L32
[3]: https://github.com/W3EDGE/w3-total-cache/blob/73911133ca843de9c20086c73aa84b134f4f9bb4/CdnEngine_S3.php#L103
[4]: https://github.com/W3EDGE/w3-total-cache/blob/73911133ca843de9c20086c73aa84b134f4f9bb4/Cdnfsd_CloudFront_Engine.php#L79
[5]: https://github.com/W3EDGE/w3-total-cache/blob/73911133ca843de9c20086c73aa84b134f4f9bb4/Cdnfsd_CloudFront_Popup.php#L409
[6]: https://github.com/W3EDGE/w3-total-cache/blob/73911133ca843de9c20086c73aa84b134f4f9bb4/Enterprise_SnsBase.php#L43
[comment]: https://wordpress.org/support/topic/iam-profile-support/#post-14588782

The primary use case for this request is when WordPress itself is running on an AWS ec2 instance or in a managed service like ECS.  In these cases the security best practice is to define AWS permissions through an [instance profile][profile] rather than require a long-lived API token to be present (`CredentialProvider::defaultProvider()` uses an instance profile when appropriate, as described [here][provider]).  Since I expect that the most common usecases for the S3 and CloudFront CDNs involve users running their entire WordPress site on AWS infrastructure, allowing users to easily opt in to using the default credential provider in this way would be a major usability win.

[profile]: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html
[provider]: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_provider.html#defaultprovider-provider

Normally when I make a request involving a relatively straightforward change like this one, I prefer to include a pull request.  However, as I haven't written PHP code in several years and am not aware of this project's standard conventions, I thought it would be simpler in this case to restrict myself to a description of the request rather than attempt to provide code that a project maintainer would likely need to substantially rewrite regardless.

jre21:
Resolves #399

If the administrator doesn't configure AWS credentials, fall back to the
default credential provider.  This is especially useful when WordPress
is running within EC2, as the default provider is compatible with the
AWS best practice of manging applications' IAM permissions through
instance profiles instead of hardcoded credentials.

### DIFF
--- a/CdnEngine_CloudFront.php
+++ b/CdnEngine_CloudFront.php
@@ -30,9 +30,13 @@ class CdnEngine_CloudFront extends CdnEngine_Base {
 			return;
 		}
 
-		$credentials = new \Aws\Credentials\Credentials(
-			$this->_config['key'],
-			$this->_config['secret'] );
+		if ( empty( $this->_config['key'] ) && empty( $this->_config['secret'] ) ) {
+			$credentials = \Aws\Credentials\CredentialProvider::defaultProvider();
+		} else {
+			$credentials = new \Aws\Credentials\Credentials(
+				$this->_config['key'],
+				$this->_config['secret'] );
+		}
 
 		$this->api = new \Aws\CloudFront\CloudFrontClient( array(
 				'credentials' => $credentials,
--- a/CdnEngine_Mirror_CloudFront.php
+++ b/CdnEngine_Mirror_CloudFront.php
@@ -29,9 +29,13 @@ class CdnEngine_Mirror_CloudFront extends CdnEngine_Mirror {
 			return;
 		}
 
-		$credentials = new \Aws\Credentials\Credentials(
-			$this->_config['key'],
-			$this->_config['secret'] );
+		if ( empty( $this->_config['key'] ) && empty( $this->_config['secret'] ) ) {
+			$credentials = \Aws\Credentials\CredentialProvider::defaultProvider();
+		} else {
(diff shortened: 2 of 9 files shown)


maxicus (approved):
Thanks for PR. Looks good.

Issue #147, opened 2020-02-29
Disallow pages cache directory into robots.txt


nigrosimone:
Sometimes bot or crawler found HTML pages into the wp-content cache directory. 

For avoid indexing of cached pages, w3tc should add a disallow rule into robots.txt of the cache directory, eg.:

```
User-agent: *
Disallow: /wp-content/cache/
```

PR #413, opened 2021-08-12
Modify robots.txt to prevent crawling of cache directory #147

jacobd91:


Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
 		return ( ( $script_data = @file_get_contents( W3TC_ADDIN_FILE_ADVANCED_CACHE ) )
 			&& strstr( $script_data, 'PgCache_ContentGrabber' ) !== false );
 	}
+
+	/**
+	 * Enables crawling blocker for cache directory
```
cssjoe, line 214: Missing full stop/period for comments.  Also, add `@since` for new methods.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+	 *
+	 * @throws Util_WpFile_FilesystemOperationException with S/FTP form if it can't get the required filesystem credentials
+	 */
+	private function robots_add_directive() {
+		$robots_path = Util_Environment::site_path() . 'robots.txt';
+
+		$robots_content = @file_get_contents( $robots_path );
+		if ( $robots_content === false )
```
cssjoe, line 222: WPCS: Use Yoda conditions and braces.
```
if ( false === $robots_content ) {
```

Did you mean to return if a `robots.txt` file was not found?  I would think that you would set the variable to an empty string.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+			return;
+
+		$new_robots_content = $this->robots_remove_from_content( $robots_content );
+		$new_robots_content = $new_robots_content . $this->robots_addon();
+
+		if ( $new_robots_content != $robots_content ) {
```
cssjoe, line 228: WPCS: Use strict comparison (`!==`).

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+			try {
+				Util_WpFile::write_to_file( $robots_path, $new_robots_content );
+			} catch ( Util_WpFile_FilesystemOperationException $ex ) {
+				throw new Util_WpFile_FilesystemModifyException(
+					$ex->getMessage(), $ex->credentials_form(),
+					'Edit file <strong>' . $robots_path .
+					'</strong> and add next lines:', $robots_path,
+					$this->robots_addon() );
```
cssjoe, line 236: WPCS: One argument per line for multi-line function/method calls.  Closing parenthesis is to be on a new line `);`.
The message to be displayed should be translatable.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+			}
+		}
+		// that file was in opcache for sure and it may take time to
+		// start execution of new modified now version
```
cssjoe, line 240: WPCS: Capitalize the first letter.  Use `/** ... */` for multi-line comments or make it one-line.  End comments with a full stop/period.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+	/**
+	 * Disables crawling blocker for cache directory
+	 *
+	 * @throws Util_WpFile_FilesystemOperationException with S/FTP form if it can't get the required filesystem credentials
+	 */
+	private function robots_remove_directive() {
+		$robots_path = Util_Environment::site_path() . 'robots.txt';
+
+		$robots_content = @file_get_contents( $robots_path );
+		if ( $robots_content === false )
```
cssjoe, line 254: WPCS: Use Yoda condition and braces.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+			return;
+
+		$new_robots_content = $this->robots_remove_from_content( $robots_content );
+		if ( $new_robots_content != $robots_content ) {
```
cssjoe, line 258: WPCS: Use strict comparison (`!==`).

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+			try {
+				Util_WpFile::write_to_file( $robots_path, $new_robots_content );
+			} catch ( Util_WpFile_FilesystemOperationException $ex ) {
+				throw new Util_WpFile_FilesystemModifyException(
+					$ex->getMessage(), $ex->credentials_form(),
+					'Edit file <strong>' . $robots_path .
+					'</strong> and remove next lines:',
+					$robots_path,  $this->robots_addon() );
```
cssjoe, line 266: WPCS: One argument per line for multi-line function/method calls.  Closing parenthesis is to be on a new line `);`.
The message to be displayed should be translatable.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+	/**
+	 * Disables crawling blocker for cache directory
+	 *
+	 * @param string  $robots_content robots.txt content
+	 * @return string
+	 * @throws Util_WpFile_FilesystemOperationException with S/FTP form if it can't get the required filesystem credentials
+	 */
+	private function robots_remove_from_content( $robots_content ) {
+		$robots_content = preg_replace(
+			"~(\r\n|\r|\n)+\\# W3TC prevent indexing of cache directory(\r\n|\r|\n)+User-agent\\: \\*(\r\n|\r|\n)+Disallow\\: \\/wp\\-content\\/cache\\/(\r\n|\r|\n)+~s",
```
cssjoe, line 287: The regex does not match the string. There are escaped backslashes and other characters that do not need to be escaped.  Maybe get the string from `robots_addon()`, replace the `*` with  `\*`, encapsulate it all with `~`, and use it as the pattern.  The `s` modifier doesn't appear to be needed.
I would have used the same approach that the plugin uses for rewrite rules; use start and end markers.  Doing so will allow more content to be added and we won't have to worry about matching all of it.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+			'', $robots_content );
```
cssjoe, line 288: WPCS: Place the closing parenthesis on its own line in multi-line calls.  Also one argument per line.

Generic_Environment.php
```diff
@@ -207,4 +209,87 @@ public function is_advanced_cache_add_in() {
...
+		/*$robots_content = preg_replace(
+			"~(\r\n|\r|\n)+\# W3TC prevent indexing of cache directory(\r\n|\r|\n)+User-agent\: \*(\r\n|\r|\n)+Disallow\: \/wp\-content\/cache\/(\r\n|\r|\n)+~s",
+			'', $robots_content );*/
```
cssjoe, line 291: Remove commented code.

cssjoe (changes requested):
Most of the items are for WordPress coding and documentation standards (WPCS).  There is a better solution for adding and removing the content block.

cssjoe (approved):

Issue #241, opened 2020-08-12
Caching of sitemap in Wordpress 5.5


superpoincare:
Wordpress 5.5 has native sitemaps. 

However, the sitemap page isn't getting cached. 

I am using disk-enhance and in the directory, I see other pages, but not this page.

mavas84:
Hello @vijayaraghavanramanan 

Thank you for taking the time to post this issue. This is a new feature introduced in WP 5.5. We need to test this and we'll get back to you once the testing is done.
Thanks!

vijayaraghavanramanan:
Worth keeping in mind that Wordpress redirects `/sitemap.xml `to `/wp-sitemap.xml `so the former shouldn't be cached.

archon810:
Any updates?

superpoincare:
Hi @mavas84,

I think it's not fixed for Apache. It does the caching but headers need to be added. 

Have commented in this thread too: https://github.com/W3EDGE/w3-total-cache/pull/416

PR #416, opened 2021-08-27
Fix for native WP sitemap caching #241

jacobd91:
Reference issue #241


maxicus:
dont see problems in the fix...
disk-enh may return .xml content as html content-type. thats controlled by pgcache.cache.nginx_handle_xml (nginx there historically, applies to both). so may be it should be enabled when testing.

well ask if i should check / test.

PR #427, opened 2021-09-14
Fix Setup Guide wizard CSS

cssjoe:
When the navigation indicator menu on the left in the Setup Guide wizard is translated, longer words were overflowing into the content area.  This fix will break long words using hyphens and wrap the text.


jacobd91 (approved):
I've tested the CSS addition and approve with no additions or changes

Issue #367, opened 2021-04-26
Provide options for how to handle missing files


timreeves:
I use nginx (exclusively) and Plesk. I am working on the Nginx config to remove all success messages from the access logs, so that monitoring them with Fail2Ban is possible (otherwise they are too large and often written, the server freezes up with race conditions). In this context I noticed while testing that when I try to access a non-existent .php file, I get the expected 404. But when I try to access any non-existent file whose type is cached by W3TC, then it receives a 301 redirect to the home page. This is due to code exported by W3TC and included in my nginx config, looking like this:

```
location ~ \.(css|htc|less|js|js2|js3|js4)$ {
    expires 36000s;
    etag on;
    if_modified_since exact;
    add_header Pragma "public";
    add_header Cache-Control "public";
    add_header X-Powered-By "W3 Total Cache/2.1.2";
    add_header Referrer-Policy "no-referrer-when-downgrade";
    try_files $uri $uri/ /index.php?$args;
}

```
See the try_files at the end: If no such file or directory exists, then the request is passed to WordPress /index.php together with the request arguments. This is the WordPress mechanism mainly used for permalink slugs and special urls provided by plugins - so when no match is found WordPress returns a 301 redirect to the domain landingpage.

But 3xx codes do not signify errors, and anyway, how would I distinguish between a 301 resulting from a probing attack and one which is bona fide, e.g. for a page which has been moved? As mentioned above, I'm keeping the logs as short as possible by suppressing all 2xx and 3xx loglines, as [demonstrated here](https://docs.nginx.com/nginx/admin-guide/monitoring/logging/#conditional). The bottom line is that we need a 403 or 404 error code at this point.

I've developed some config which I can put before the code block from W3TC is included:

```
# Catch attempts by nasty bots to access executables which are not used on the server
location ~ \.(js2|js3|js4|csh|exe|jar|pl|py|pyo|pyc|rb|sh)$ { deny all; return 444; }

# Handle JS ourselves, i.e. pre-empt W3TC on this one to get a 403 in the access log
# 403 if no file found because we come down hard on IPs causing 403's - and this is JS...

location ~ \.js$ {
    expires 36000s;
    etag on;
    if_modified_since exact;
    add_header Pragma "public";
    add_header Cache-Control "public";
    add_header Referrer-Policy "no-referrer-when-downgrade";
    try_files $uri $uri/ =403;
}
```
[...]

PR #428, opened 2021-09-14
Expanded browsercache options for 404 nginx handling #367

jacobd91:
#367


BrowserCache_ConfigLabels.php
```diff
@@ -7,7 +7,11 @@ public function config_labels( $config_labels ) {
 				'browsercache.enabled' => __( 'Browser Cache:', 'w3-total-cache' ),
 				'browsercache.replace.exceptions' => __( 'Prevent caching exception list:', 'w3-total-cache' ),
 				'browsercache.no404wp' => __( 'Do not process 404 errors for static objects with WordPress', 'w3-total-cache' ),
-				'browsercache.no404wp.exceptions' => __( '404 error exception list:', 'w3-total-cache' ),
+				'browsercache.no404wp.response.cssjs' => __( '404 replacement response code for CSS/JS mime types', 'w3-total-cache' ),
+				'browsercache.no404wp.response.html' => __( '404 replacement response code for HTML mime types', 'w3-total-cache' ),
+				'browsercache.no404wp.response.other' => __( '404 replacement response code for Other mime types', 'w3-total-cache' ),
+				'browsercache.no404wp.response.exceptions' => __( '404 replacement response code for Exceptions URIs', 'w3-total-cache' ),
+				'browsercache.no404wp.exceptions' => __( '404 exception list:', 'w3-total-cache' ),
```
cssjoe, line 14: You don't need to change old code if it's not relevant.

BrowserCache_Environment_Nginx.php
```diff
@@ -131,12 +127,14 @@ public function generate( $mime_types, $cdnftp = false ) {
 
 			$impoloded = implode( '|', $exceptions );
 			if ( !empty( $impoloded ) ) {
-				$wp_uri = network_home_url( '', 'relative' );
-				$wp_uri = rtrim( $wp_uri, '/' );
+				$file_code = rtrim( network_home_url( '', 'relative' ), '/' ) . '/index.php?$args;';
+
+				if( $this->c->get_integer( 'browsercache.no404wp.response.exceptions' ) ){
```
cssjoe, line 132: WPCS: Add spaces for control structures. -- `if ( ... ) {`

BrowserCache_Environment_Nginx.php
```diff
@@ -131,12 +127,14 @@ public function generate( $mime_types, $cdnftp = false ) {
...
+					$file_code = "=" . $this->c->get_integer( 'browsercache.no404wp.response.exceptions' );
```
cssjoe, line 133: Use single quotes for strings that do not need character translation or escaping other single quotes -- `'='`.

BrowserCache_Environment_Nginx.php
```diff
@@ -302,19 +296,19 @@ private function generate_section( &$rules, $mime_types, $section ) {
 				$this->c, $section );
 			$rules .= '    ' . implode( "\n    ", $subrules ) . "\n";
 
-			if ( !$this->c->get_boolean( 'browsercache.no404wp' ) ) {
-				$wp_uri = network_home_url( '', 'relative' );
-				$wp_uri = rtrim( $wp_uri, '/' );
+			if ( $this->c->get_boolean( 'browsercache.no404wp' ) ) {
```
cssjoe, line 299: It looks like this condition was negated.  If `browsercache.no404wp` is true, then we should not be trying to rewrite the request through WordPress's `index.php` file.

cssjoe (changes requested):
See file comments.  Also, the 404 handling affects Apache htaccess or Nginx conf.  The new features should be consistent with that.

BrowserCache_ConfigLabels.php
jacobd91: I don't follow. The only old code here I changed was the exceptions label for better consistency with the new options and a few other wording changes to hopefully avoid convusion

BrowserCache_Environment_Nginx.php
jacobd91: I believe the original condition was actually incorrect as the generate_section is called for each mime type. If this condition were left as a !condition it wouldn't apply the "no 404" functionality. I will retest and confirm this

BrowserCache_ConfigLabels.php
cssjoe: It's the change from "404 error exception list" to "404 exception list".

maxicus (changes requested):
extremely risky.

such change will break things with almost 100% probability.
many changes unrelated to issue itself.

i see at least 2 conditions "just removed"

maxicus:
When no file present - it's passed to WP by default (that's what original WP rules do) and WP decides what to respond.
In a case of those blog his WP decided to return 301 - it's choice and it's how it should be. It's not W3TC's decision.
=> behavior described in original ticket is "expected and correct".
=> seems whole fix is just not needed.

cssjoe:
Brad, Max, and I decided that this feature is not worth the risk of causing bigger issues by making more complex rewrite rules.

PR #434, opened 2021-09-28
Update Cdn_Plugin.php

hanicker:
Execute purge on cdn only if hosting is enabled, to prevent unneded delays

### DIFF
--- a/Cdn_Plugin.php
+++ b/Cdn_Plugin.php
@@ -235,7 +235,9 @@ class Cdn_Plugin {
 
 		$cdn_engine = $this->_config->get_string( 'cdn.engine' );
 		if ( Cdn_Util::is_engine_mirror( $cdn_engine ) ) {
-			$common->purge( $files, $results );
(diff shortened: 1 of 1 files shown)


maxicus (approved):

Cdn_Plugin.php
```diff
@@ -235,7 +235,9 @@ function update_attachment_metadata( $metadata ) {
 
 		$cdn_engine = $this->_config->get_string( 'cdn.engine' );
 		if ( Cdn_Util::is_engine_mirror( $cdn_engine ) ) {
-			$common->purge( $files, $results );
+			if ($this->_config->get_string( 'cdn.uploads.enable' )) {
```
cssjoe, line 238: The setting is a boolean and not a string.  You can use the following:
```
if ( $this->_config->get_boolean( 'cdn.uploads.enable' ) ) {
```

cssjoe (changes requested):

cssjoe (approved):

PR #465, opened 2021-11-30
Cloudflare enhancement to allow for either global API key or tokens

jacobd91:


fritexvz:
Kindly, may I ask, as far as the user chooses to use a Cloudflare API Token - which can be restricted by the needed/required permissions, may I ask will somewhere be stated which permissions does the user need to allow per the API token so W3TC plugin would continue working properly (at least in terms of Purge the Cache, purge the cache upon post modification of the current settings available)?

Could be I misread them if they are already stated for future release of W3TC plugin in some help/setup process while configuring/connecting Cloudflare extension from the W3TC interface.

Reference: 
https://api.cloudflare.com/#getting-started-endpoints

Image from Cloudflare Dashboard -> API Tokens:

api_token

Nevertheless, we could limit the API token with required permissions per zone/domain, that way it wouldn't have permissions to view/edit others from the Cloudflare account.

Not to mention the fact that if the server got compromised, the E-mail and Global API Key from the Cloudflare Account are being stored into the plaintext JSON formatted "master.php" file of the "w3tc-config" directory, right?
This is the main issue if it happens.

It would be much better to advise W3TC users who are using W3TC plugin with the Cloudflare extension enabled to generate API Token per each zone/domain with the required permissions that W3TC plugin w/ Cloudflare extension needs to work properly, just in case.

Kindly, thank you very much for your time and effort making this one!

maxicus (approved):

cssjoe (approved):

mobiusint:
Hi there, so sorry to bump this but I can't get the token to work with Cloudflare and email address. It says invalid headers. Any idea how to get this to work so we don't have to install cloudflare plugin?

PR #467, opened 2021-12-01
Util_File.php fix for mkdir calls not pre-checking if the provided path is a file

jacobd91:
It was only checking if the provided path was a directory or not


cssjoe (changes requested):
It's cleaner to use `file_exists()`.
https://www.php.net/manual/en/function.file-exists.php
>Checks whether a file or directory exists.

cssjoe (approved):

PR #468, opened 2021-12-01
Fix Image Service responses and handle API rate-limit

cssjoe:
This fixes response messages when the API server reports that a rate-limit (hourly or monthly) has been exceeded for image submissions.


jacobd91 (approved):
I see no issues with these changes and they look straightforward and simple. Approved

PR #472, opened 2021-12-02
Check filesystem and provide better messaging

cssjoe:
This can be tested by changing the defined WP_Filesystem type in `wp-config.php` using **one** of the following definitions::
```
define( 'FS_METHOD', 'direct' );
define( 'FS_METHOD', 'ssh2' );
define( 'FS_METHOD', 'ftpext') ;
define( 'FS_METHOD', 'ftpsockets' );
```


jacobd91 (approved):
Tested and approved

PR #474, opened 2021-12-16
Fix access log test

cssjoe:
Reference: https://wordpress.org/support/topic/access-logs-minimum-permissions/


jacobd91 (approved):
Approved

PR #478, opened 2022-01-07
Add visibility option.  Fixes #476.

### LINKED ISSUE #476: Add option for Image Service WP_Query behavior
There is a WP_Query filter used when the plugin is active, which hides the WebP image attachments created by the Image Service.  At least one user reported that their queries were slower even when the Image Service extension is disabled.

We'll add an option to choose the behavior of the filter.
For example: Show converted images in Media Library
* Never
* If the Image Service extension is active
* Always

See https://github.com/W3EDGE/w3-total-cache/blob/2.2.1/Root_Loader.php#L166-L211

cssjoe:
An option is added to choose the behavior of hiding converted image attachments via WP_Query filters.  By default, the converted images are never shown, unless the plugin is deactivated.  The new option allows them to be shown all of the time or only when the Image Service extension is active.

### DIFF
--- a/ConfigKeys.php
+++ b/ConfigKeys.php
@@ -2499,6 +2499,7 @@ $keys = array(
 		'default' => array(
 			'compression' => 'lossy',
 			'auto'        => 'enabled',
+			'visibility'  => 'never',
 		),
 	),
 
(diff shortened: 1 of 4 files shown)


jacobd91 (approved):
Looks straightforward and clean. Approved

PR #486, opened 2022-02-11
Remove deprecated opcache flush

cssjoe:
From wporg feedback.


maxicus (approved):
im fine. afaik noone use it atm.

PR #488, opened 2022-02-16
Part for WP feedback

cssjoe:


Util_Request.php
```diff
@@ -23,14 +23,10 @@ static function get( $key, $default = null ) {
 			$value = $request[$key];
 
 			if ( defined( 'TEMPLATEPATH' ) ) {
-				$value = Util_Environment::stripslashes( $value );
+				return sanitize_text_field( Util_Environment::stripslashes( $value ) );
```
jacobd91, line 26: If the $value is an array this won't properly sanitize the array value(s) unless you utilize something like map_deep or array_map such as that on line 29 and line 1168 in Util_Environment for function stripslashes respectively

jacobd91 (changes requested):

Util_Request.php
jacobd91: return is_array( $value ) ? map_deep( Util_Environment::stripslashes( $value ), 'sanitize_text_field' ) : sanitize_text_field( $value );

jacobd91: Or keep the original as $value assignment and allow the 29 return to process the sanitation

jacobd91 (approved):
Approved

Issue #67, opened 2019-10-23
Add TLS support for Redis


mavas84:
Reference topic: https://wordpress.org/support/topic/add-tls-support-for-redis/

maxicus:
implemented by https://github.com/W3EDGE/w3-total-cache/commit/b31235f81d5ff7875e55a8939ba015014bafd26c

MuffintopSpeedos:
Was this feature removed? 

I'm getting log spam of "[php7:notice] [pid 302932] [client <my IP>:56640] read error on connection to <redis host>:25061"

PR #494, opened 2022-04-13
Redis: support connecting to servers using TLS with self-signed certificates (e.g., Heroku Redis)

dbertram:
Adds on to https://github.com/W3EDGE/w3-total-cache/issues/67 to include the option of _not_ verifying the redis server's TLS certificate when connecting over TLS.

This option is needed when connecting to a redis server configured to use TLS with self-signed certificates. In particular, [Heroku Redis](https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-php) falls under this category.

With this new option TLS certificates continue to be verified by default, but users who need to now have the option of disabling certificate verification.

While working on this, I noticed [this comment from the PR where TLS support was added](https://github.com/W3EDGE/w3-total-cache/issues/67#issuecomment-828548567), so I also updated the UI string to include a hint about how to use redis over TLS.

### DIFF
--- a/languages/w3-total-cache.pot
+++ b/languages/w3-total-cache.pot
@@ -7585,9 +7585,15 @@ msgstr ""
 msgid "symbols (set to 0 to disable)"
 msgstr ""
 
-#: inc/options/parts/memcached_extension.php:23
-#: inc/options/parts/redis_extension.php:25 inc/options/parts/memcached.php:22
+#: inc/options/parts/redis_extension.php:25
 #: inc/options/parts/redis.php:23
+msgid ""
+"Multiple servers may be used and seperated by a comma; e.g. 192.168.1.100:"
+"11211, domain.com:22122. To use TLS, prefix server with tls://"
+msgstr ""
+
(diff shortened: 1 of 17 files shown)


languages/w3-total-cache.pot
```diff
@@ -7668,6 +7674,11 @@ msgstr ""
 
 #: inc/options/parts/redis.php:30
 msgid ""
+"Verify the server\'s certificate when connecting via TLS."
+msgstr ""
+
+#: inc/options/parts/redis.php:37
+msgid ""
```
dbertram, line 7681: Full disclosure: I'm not familiar with how localization is set up for this plugin and I couldn't find any developer documentation that mentioned it, so I made the changes to this file by imitating what I saw elsewhere. 😅

maxicus (approved):
it's silly to use tls and not use tls at the same time, but nvm

cssjoe (approved):

cssjoe:
@dbertram Thank you for your contribution.  This will be included in the next release (2.2.2).

PR #500, opened 2022-05-03
Escaping translation

jacobd91:


cssjoe:
I moved the changes to https://github.com/W3EDGE/w3-total-cache/pull/502

Issue #506, opened 2022-05-18
`db-cluster-config.php` does not work for non network mode


jpSimkins:
I have a single "blog" website. I am NOT using "network mode". I was instructed when I bought this that this still works if I put the `db-cluster-config.php` in the `wp-content` location. I have done this and this does not work.

I have already set the `define('W3TC_PRO_DEV_MODE', true);` in my `wp-config.php` and set the license in the setting page.

I am testing on my local DEV since I am already using HyperDB to have a multi-node deployment of WordPress running in Fargate on PROD. I have a local DB cluster setup for DEV too. I need to verify this will work on DEV before I even consider putting this on PROD.

I have removed HyperDB and added the drop-in for W3TC `db.php` and placed the `db-cluster-config.php`. I can verify through debugging that I am hitting the `db.php` drop-in as expected. I am never hitting the `db-cluster-config.php` file in debug.

Support told me:
> As for the DB cluster, initially, this was intended to network mode, however, it can be used in the single-blog setups, but the mentioned check-box will not be visible, however, once the dbcluster-config-sample.php file is edited and  placed in /wp-content/db-cluster-config.php - it starts to work

But this seems to be missing that the `db.php` needs to be in place as well. Which is obvious but was not mentioned. The only way the `db-cluster-config.php` is loaded, is when the `db.php` file is present. If you remove it, it will not load `db-cluster-config.php`.

So I have the files in place. I added this to the `db-cluster-config.php`:
```php
if (defined('DB_READ_HOST')) {
    // Overrides the databases key above
    $w3tc_dbcluster_config['databases'] = array(
        'master' => array(
            'host'     => DB_HOST,
            'user'     => DB_USER,
            'password' => DB_PASSWORD,
            'name'     => DB_NAME,
            'write'    => true,
            'read'     => false,
        ),
        'slave' => array(
            'host'     => DB_READ_HOST,
            'user'     => DB_USER,
            'password' => DB_PASSWORD,
            'name'     => DB_NAME,
            'write'    => false,
            'read'     => true,
            'timeout'  => 0.3,
        ),
    );
}
```
I modified this since the commented out sections replaced `$w3tc_dbcluster_config` which would result in the other indexes being removed. e.g. `persistent`, `check_tcp_responsiveness`, ...

I tested with original uncommitted code and my changes but it's never worked as this file is never processed.

I do have `DB_READ_HOST` defined but this still isn't working. This is the same config I used with HyperDB so it should work but I get nothing.

[...]

PR #507, opened 2022-05-20
fix: Database cluster logic

jpSimkins:
- Solved issue with db-cluster-config.php not loading
- Fixed bug with using object variables in a string when setting charset and collation
- Fixed broken pagination due to using incorrect mysqli_ method

See bug report: https://github.com/W3EDGE/w3-total-cache/issues/506

### DIFF
--- a/Enterprise_Dbcache_WpdbInjection_Cluster.php
+++ b/Enterprise_Dbcache_WpdbInjection_Cluster.php
@@ -122,11 +122,13 @@ class Enterprise_Dbcache_WpdbInjection_Cluster extends DbCache_WpdbInjection {
 		global $wpdb_cluster;
 		$wpdb_cluster = $this;
 
-		if ( isset( $GLOBALS['w3tc_dbcluster_config'] ) ) {
-			$this->apply_configuration( $GLOBALS['w3tc_dbcluster_config'] );
-		} elseif ( file_exists( WP_CONTENT_DIR . '/db-cluster-config.php' ) ) {
+		if ( !isset( $GLOBALS['w3tc_dbcluster_config'] ) && file_exists( WP_CONTENT_DIR . '/db-cluster-config.php' ) ) {
 			// The config file resides in WP_CONTENT_DIR
 			require WP_CONTENT_DIR . '/db-cluster-config.php';
+		}
+
+		if ( isset( $GLOBALS['w3tc_dbcluster_config'] ) ) {
+			$this->apply_configuration( $GLOBALS['w3tc_dbcluster_config'] );
 		} else {
 			$this->_reject_reason = 'w3tc dbcluster configuration not found, ' .
 				'using single-server configuration';
@@ -575,7 +577,7 @@ class Enterprise_Dbcache_WpdbInjection_Cluster extends DbCache_WpdbInjection {
 			if ( preg_match( '/^\s*SELECT\s+SQL_CALC_FOUND_ROWS\s/i', $query ) ) {
 				if ( false === strpos( $query, "NO_SELECT_FOUND_ROWS" ) ) {
 					$this->wpdb_mixin->timer_start();
-					$this->wpdb_mixin->_last_found_rows_result = mysqli_query( $this->wpdb_mixin->dbh, "SELECT FOUND_ROWS()" );
+					$this->wpdb_mixin->_last_found_rows_result = mysqli_num_rows( $this->wpdb_mixin->dbh, "SELECT FOUND_ROWS()" );
 					$elapsed += $this->wpdb_mixin->timer_stop();
 					++$this->wpdb_mixin->num_queries;
 					$query .= "; SELECT FOUND_ROWS()";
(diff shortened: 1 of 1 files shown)


Enterprise_Dbcache_WpdbInjection_Cluster.php
```diff
@@ -122,11 +122,13 @@ function initialize() {
 		global $wpdb_cluster;
 		$wpdb_cluster = $this;
 
-		if ( isset( $GLOBALS['w3tc_dbcluster_config'] ) ) {
-			$this->apply_configuration( $GLOBALS['w3tc_dbcluster_config'] );
-		} elseif ( file_exists( WP_CONTENT_DIR . '/db-cluster-config.php' ) ) {
+		if ( !isset( $GLOBALS['w3tc_dbcluster_config'] ) && file_exists( WP_CONTENT_DIR . '/db-cluster-config.php' ) ) {
```
jpSimkins, line 125: This is needed otherwise the `db-cluster-config.php` file is never loaded.

Enterprise_Dbcache_WpdbInjection_Cluster.php
```diff
@@ -874,9 +876,9 @@ function _db_connect_fallback() {
 		if ( !mysqli_select_db( $this->wpdb_mixin->dbh, DB_NAME ) )
 			return $this->wpdb_mixin->bail( "We were unable to select the database." );
 		if ( !empty( $this->wpdb_mixin->charset ) ) {
-			$collation_query = "SET NAMES '$this->wpdb_mixin->charset'";
+			$collation_query = "SET NAMES '{$this->wpdb_mixin->charset}'";
```
jpSimkins, line 879: This is the correct way to handle object variables in a string. This will throw a PHP error otherwise.

Enterprise_Dbcache_WpdbInjection_Cluster.php
```diff
@@ -575,7 +577,7 @@ function query( $query ) {
 			if ( preg_match( '/^\s*SELECT\s+SQL_CALC_FOUND_ROWS\s/i', $query ) ) {
 				if ( false === strpos( $query, "NO_SELECT_FOUND_ROWS" ) ) {
 					$this->wpdb_mixin->timer_start();
-					$this->wpdb_mixin->_last_found_rows_result = mysqli_query( $this->wpdb_mixin->dbh, "SELECT FOUND_ROWS()" );
+					$this->wpdb_mixin->_last_found_rows_result = mysqli_num_rows( $this->wpdb_mixin->dbh, "SELECT FOUND_ROWS()" );
```
jpSimkins, line 580: This fixes pagination not working as this is the only way to properly fetch `SELECT FOUND ROWS()`

jpSimkins: Actually, this worked but was still incorrect. I pushed a new update to solve this.

Enterprise_Dbcache_WpdbInjection_Cluster.php
```diff
@@ -575,7 +577,7 @@ function query( $query ) {
 			if ( preg_match( '/^\s*SELECT\s+SQL_CALC_FOUND_ROWS\s/i', $query ) ) {
 				if ( false === strpos( $query, "NO_SELECT_FOUND_ROWS" ) ) {
 					$this->wpdb_mixin->timer_start();
-					$this->wpdb_mixin->_last_found_rows_result = mysqli_query( $this->wpdb_mixin->dbh, "SELECT FOUND_ROWS()" );
+					$this->wpdb_mixin->_last_found_rows_result = mysqli_num_rows( $this->wpdb_mixin->result );
```
cssjoe, line 580: The result should be a `mysqli_result` and not a `string|int`, so `mysqli_query` is the correct function.  Lines 563 - 564 check if the query is "SELECT FOUND_ROWS()" and uses the cached result instead of repeating the query.

cssjoe (changes requested):
@jpSimkins Thank you for your contribution.

The result should be a `mysqli_result` and not a `string|int`, so `mysqli_query` is the correct function.  Lines 563 - 564 check if the query is "SELECT FOUND_ROWS()" and uses the cached result instead of repeating the query.  Lines 612, 613, and 619 expect the value to be a `mysqli_result`.

The other changes look good.

Enterprise_Dbcache_WpdbInjection_Cluster.php
jpSimkins: Looking into this a bit more, I don't see any reason this needs to be a `mysqli_query` result. The only area this affects is: `is_resource( $this->wpdb_mixin->_last_found_rows_result )` which can be updated to work with `mysqli_num_rows`. I will work on this now as this is necessary to get database cluster to work as it is currently broken (no pagination).

jpSimkins: Looking into this a bit more, I don't see any reason this needs to be a `mysqli_query` result. The only area this affects is: `is_resource( $this->wpdb_mixin->_last_found_rows_result )` which can be updated to work with `mysqli_num_rows`. I will work on this now as this is necessary to get database cluster to work as it is currently broken (no pagination).

jpSimkins: ok, so the underlying issue is that `is_resource( $this->wpdb_mixin->_last_found_rows_result )` is always returning `false`. This is due to `mysqli` not using a resource, it's using objects (`mysql` legacy probably`). This now fixes pagination and still uses the `mysqli_query` which is desired.

Enterprise_Dbcache_WpdbInjection_Cluster.php
```diff
@@ -559,7 +561,7 @@ function query( $query ) {
 		$this->wpdb_mixin->last_query = $query;
 
 		if ( preg_match( '/^\s*SELECT\s+FOUND_ROWS(\s*)/i', $query )
-			&& is_resource( $this->wpdb_mixin->_last_found_rows_result ) ) {
+			&& is_object( $this->wpdb_mixin->_last_found_rows_result ) ) {
```
cssjoe, line 564: This should work.  A stricter check would be to use `is_a( $this->wpdb_mixin->_last_found_rows_result, 'mysqli_result' )`, although `is_object()` is ok.

cssjoe (approved):
@jpSimkins Thank you for the fix.  We'll include this in our next patch release 2.2.2.

Issue #508, opened 2022-05-21
DB Cache causes infinite loop when using `db-cluster-config.php` in non-network mode (not sure if it does in network mode)


jpSimkins:
Hello,

After I fixed the issues with https://github.com/W3EDGE/w3-total-cache/issues/506, I turned on DB caching (to disk for testing) and the site is in an infinite loop. This happened as soon as I submitted the settings. It saved the setting and the next page load it is crashing the site (frontend and back).

```
( ! ) Error: Maximum function nesting level of '200' reached, aborting! in /var/www/html/wp-content/plugins/w3-total-cache/Config.php on line 207
--
1 | 0.0001 | 366168 | {main}( ) | .../index.php:0
2 | 0.0002 | 366448 | require( '/var/www/html/wordpress/wp-blog-header.php' ) | .../index.php:17
3 | 0.0002 | 367056 | require_once( '/var/www/html/wordpress/wp-load.php' ) | .../wp-blog-header.php:13
4 | 0.0002 | 367720 | require_once( '/var/www/html/wp-config.php' ) | .../wp-load.php:55
5 | 0.0003 | 373256 | require_once( '/var/www/html/wordpress/wp-settings.php' ) | .../wp-config.php:107
6 | 0.0052 | 1390552 | wp_not_installed( ) | .../wp-settings.php:159
7 | 0.0052 | 1390552 | is_blog_installed( ) | .../load.php:755
8 | 0.0052 | 1390960 | wp_load_alloptions( ) | .../functions.php:1749
9 | 0.0052 | 1391072 | W3TC\DbCache_WpdbNew->get_results( ) | .../option.php:287
10 | 0.0052 | 1391200 | W3TC\DbCache_WpdbNew->query( ) | .../wp-db.php:2829
11 | 0.0052 | 1391200 | W3TC\DbCache_WpdbInjection_QueryCaching->query( ) | .../DbCache_WpdbNew.php:122
12 | 0.0053 | 1391232 | W3TC\DbCache_WpdbInjection_QueryCaching->_can_cache( ) | .../DbCache_WpdbInjection_QueryCaching.php:115
13 | 0.0053 | 1391232 | W3TC\DbCache_WpdbInjection_QueryCaching->_can_cache_once_per_request( ) | .../DbCache_WpdbInjection_QueryCaching.php:407
[...]

PR #509, opened 2022-05-21
fix: DB Cache infinite loop when using `db-cluster-config.php`

jpSimkins:
- Fix DB Cache causes infinite loop when using db-cluster-config.php in non-network mode

See: https://github.com/W3EDGE/w3-total-cache/issues/508


cssjoe (approved):
@jpSimkins Thank you for your contribution.  We used `esc_url_raw` in another branch.  We are ok with using `filter_var` it that worked better for you.  The 2.2.2 patch release will be released soon after we complete more testing.  Please note that the `master` branch is currently being reworked for escaping and sanitization.

PR #518, opened 2022-06-14
2.2.2-rc.1

cssjoe:
Patch release candidate 2.2.2-rc.1

### DIFF
--- a/Cache_Redis.php
+++ b/Cache_Redis.php
@@ -505,6 +505,18 @@ class Cache_Redis extends Cache_Base {
 						);
 					}
 
+					// Catch PHP errors.
+					set_error_handler( // phpcs:ignore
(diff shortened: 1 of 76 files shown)


bwmarkle (approved):

KZeni (commented):
This appears to be tied to https://github.com/W3EDGE/w3-total-cache/issues/523 with 2.2.2 having issues being spread to more sites as they update.

Could this (seemingly being able to revert Cache_Redis.php back to what it was in 2.2.1 without issue) be reverted back as a hotfix to prevent more sites from encountering the issues sooner than later while this can be revisited for another future release after that?

Some sites are going down entirely due to PHP warnings becoming a fatal server error with Redis enabled so even updating to a patched version could be problematic for some if they update before 2.2.2 is recalled or patched to quickly prevent this.

Issue #515, opened 2022-06-01
ErrorException: Trying to get property 'post_type' of non-object


mavas84:
The problem occurs randomly when the ACF fields are edited.
Reference topic: https://wordpress.org/support/topic/trying-to-get-property-post_type-of-non-object-11/

Striffly:
This occurs for ACf fields, but also with menu.
Steps to reproduce : 
- Bedrock + Sage installation, with W3 Total Cache composer install
- Use **only** page caching with Redis
- Add page to menu
- Error occured

![image](https://user-images.githubusercontent.com/13284087/175033019-6196b32e-9f99-413b-bfdc-1ff1c1aafdf8.png)

PR #530, opened 2022-07-04
:bug: PgCache : Check if post is not empty before flush

Striffly:
Related issue : https://github.com/W3EDGE/w3-total-cache/issues/515


cssjoe (approved):
@Striffly Thank you for your contribution.  I added formatting for WordPress Coding standards.

PR #532, opened 2022-07-05
Avoid PHP warning about pconnect() expecting at most 5 parameters with 6 given on some setups (likely using an older phpredis)

KZeni:
I'm currently getting:
```
PHP Warning:  Redis::pconnect() expects at most 5 parameters, 6 given
```
on a server running PHP 7.2 with phpredis 3.1.1 (via php72-pecl-redis).

I'm thinking this is due to phpredis introducing the additional parameters only in later versions. As such, this now version checks for phpredis newer than 4 (so all of v5 and newer) to use what W3TC already had while it reduces the parameters by 1 if it's an older version.

Of course, this can be updated/adapted to be more granular (ex. expand on this to add context support for that specific phpredis version where that support was added) and/or be implemented in a better way overall.

I've also posted this on the WP.org support forums here: https://wordpress.org/support/topic/php-warning-about-pconnect-expecting-at-most-5-parameters-with-6-given/
As well as being a GitHub issue here: https://github.com/W3EDGE/w3-total-cache/issues/533 (just keeping things tied together.)

### DIFF
--- a/Cache_Redis.php
+++ b/Cache_Redis.php
@@ -474,32 +474,22 @@ class Cache_Redis extends Cache_Base {
 				$server   = $this->_servers[ $index ];
 				$accessor = new \Redis();
 
-				$phpredis_version = phpversion('redis');
-				if ($phpredis_version) {
-					$phpredis_version = (int) $phpredis_version[0]; // Only care about the major version
-				} else {
-					$phpredis_version = 0;
-				}
-				if ($phpredis_version > 4) {
-					$phpredis_modern = true;
-				} else {
-					$phpredis_modern = false;
-				}
+				$phpredis_modern = version_compare( phpversion( 'redis' ), '5', '>=' );
 
 				if ( substr( $server, 0, 5 ) === 'unix:' ) {
 					if ( $this->_persistent ) {
-						if ($phpredis_modern) {
+						if ( $phpredis_modern ) {
 							$accessor->pconnect(
-								trim(substr($server, 5)),
+								trim( substr( $server, 5 ) ),
 								null,
 								$this->_timeout,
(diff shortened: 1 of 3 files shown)


KZeni:
As an aside, I needed to come up with this since I was getting my server log filled with that pconnect parameter count warning all of a sudden (seems to have coincided with W3TC v2.2.2 and/or v2.2.3.) This hopefully helps prevent the issue rather than simply turning off the error_log being used for Redis commands as a means to address this.

This PR is currently largely untested & should be reviewed.

KZeni:
Additionally, I imagine the `connect` function (not just `pconnect`) could/should be updated to follow suit (not familiar with phpredis to say one way or the other). `pconnect` was just the one I was encountering this issue with, but it could affect both where both could be set up in a similar fashion.

I do plan on moving these sites to a newer setup where you can provide more than 5 parameters for that function, but I imagine there might be others out there that haven't updated from PHP 7.x or so (or some other reason an older version of phpredis might be used) where this warning being logged on effectively every Redis pconnect request has server logs absolutely ballooning in size to an unreasonable degree (definitely unlike what pre-2.2.2 had with everything else about the site/server remaining the same.)

cssjoe:
@KZeni Thank you for your contribution.  Please review and test the additional changes that I made to your pull request and let us know if everything is working as expected.  I used version_compare() and addressed the non-persistent connections as well.  I also conditionally offered the Redis `read_timeout` setting on plugin settings pages.

KZeni:
@cssjoe Those changes look to be good. Also, I've tested it across multiple sites, and the issue is looking to be resolved with this latest version of things. Thanks for the quick response on this!

PR #542, opened 2022-08-02
Fix Redis undefined array key warnings

michaelw85:
For details see: https://wordpress.org/support/topic/undefined-array-key-timeout/#post-15878314

TLDR;
`timeout`, `retry_interval` and `read_timeout` are missing when config is passed here: https://github.com/W3EDGE/w3-total-cache/blob/master/Cache_Redis.php#L99


cssjoe (approved):

cssjoe:
@michaelw85 Thank you for your contribution.  We'll include this fix in the next patch release (2.2.4).

PR #543, opened 2022-08-02
Woocommerce Variation Image Gallery plugin CDN fix

jacobd91:
This fix will apply to any users utilizing W3TC and any plugin/theme that uses wp_get_attachment_image_src with CDN enabled.

You'll need [WooCommerce](https://wordpress.org/plugins/woocommerce/) and [Variation Images Gallery for WooCommerce](https://wordpress.org/plugins/woo-product-variation-gallery/).  You'll also need a variable product.

Once you have WooCommerce and product in place, you should be able to toggle the "Host attachments" setting (on `wp-admin/admin.php?page=w3tc_cdn`) and the product page should serve local/CDN accordingly.  If my fix is working then when "Host attachments" is enabled all product images should be served via the CDN assuming they're uploaded to the bucket.

### DIFF
--- a/Cdn_Plugin.php
+++ b/Cdn_Plugin.php
@@ -754,14 +754,14 @@ class Cdn_Plugin {
 	 * @return 	array
 	 */
 	function wp_get_attachment_image_src( $image ) {
-		$url = trim( $image[0] );
+		$url = ( ! emtpy( $image[0] ) ? trim( $image[0] ) : null );
 		
 		if ( ! empty( $url ) ) {
 			$parsed          = parse_url( $url );
 			$uri             = ( isset( $parsed['path'] ) ? $parsed['path'] : '/' ) . ( isset( $parsed['query'] ) ? '?' . $parsed['query'] : '' );
 			$wp_upload_dir   = wp_upload_dir();
 			$upload_base_url = $wp_upload_dir['baseurl'];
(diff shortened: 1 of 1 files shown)


cssjoe (changes requested):

Cdn_Plugin.php
```diff
@@ -743,6 +746,33 @@ private function wp_prepare_attachment_for_js_url( $url ) {
...
 
+	/**
+	 * Adjusts attachment image src to cdn. This is for those who rely on
+	 * wp_get_attachment_image_src()
+	 *
+	 * @param 	array   $image	Image object
+	 * @return 	array
+	 */
+	function wp_get_attachment_image_src( $image ) {
+		$url = ( ! emtpy( $image[0] ) ? trim( $image[0] ) : null );
```
cssjoe, line 757: Fix typo.  Also, it;s better practice to evaluate true instead of not (`!`).  Also, parentheses are not needed around the ternary.
Might as well use: `$url = empty( $image[0] ) ? null : trim( $image[0] );`

cssjoe (changes requested):

Cdn_Plugin.php
jacobd91: Updated to your suggestion. Sorry for missing my typo before submitting

cssjoe (approved):

PR #544, opened 2022-08-04
Add setting for AWS S3 public objects in ACL

cssjoe:
To test, configure CDN using AWS S3, and then upload media.  The public option is enabled by default in the settings.  If enabled, the uploaded media will be publicly accessible via a bucket URL.  Your site should be using the CDN URLs on pages.


jacobd91 (changes requested):

jacobd91 (approved):
Approved!

Issue #535, opened 2022-07-07
Cloudfront Invalidation charges notification - Enable Manual purging by default for AWS


mavas84:
Cloudflare doesn't charge for "request", but for invalidation path.
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html#PayingForInvalidation
The first 1,000 invalidation paths that you submit per month are free; you pay for each invalidation path over 1,000 in a month. An invalidation path can be for a single file (such as `/images/logo.jpg`) or for multiple files (such as `/images/*`). A path that includes the `*` wildcard counts as one path even if it causes CloudFront to invalidate thousands of files.
What can be done:
invalidate all your images per-case basis,
i.e.
```
https//my-host/wp-content/uploads/2022/01/my-jan-image1.jpg
https//my-host/wp-content/uploads/2022/01/my-jan-image2.jpg
https//my-host/wp-content/uploads/2022/02/my-feb-image1.jpg
```
That will be invalidated as 3 paths and that's what w3tc does.
2. invalidate everything.
That's what w3tc's purge all button does. All objects will be removed from the cache the traffic will grow and performance temporarily decline. Also, of course, this needs to be done once after multiple posts are updated, otherwise, if you purge after each change you'll get a worse situation in the end (3 invalidate-all requests for 3 posts).
3. invlidate grouped based on knowledge what you've modified. in the example below you may invalidate one of:
3.1 https//my-host/wp-content/uploads/2022/*  - one path, will invalidate all 2022 images only
3.2 https//my-host/wp-content/uploads/2022/01/*, https//my-host/wp-content/uploads/2022/02/* - 2 paths, but will invalidate only all 2022 jan and feb images. 
![image](https://user-images.githubusercontent.com/40050744/177826355-2c097704-663d-473f-9e75-2333535ddcfb.png)

The `Only Purge CDN Manually `checkbox was added specifically because of this problem in the `Performance>CDN>Advanced` section. Once checked, any changes are not invalidated automatically, so the cache needs to be purged manually as explained above.
This is not, however, enabled by default. 
This checkbox should be enabled by default if one of the following CDN types are selected in the Performance>General Settings>CDN section under CDN type:
```
Origin Pull: Amazon Cloudfront
Origin push: Amazon Cloudfront over S3
```
Also, a detailed explanation about the Cloudfron invalidation and excessive charges should've been added in the description under the `Only Purge CDN Manually `checkbox

PR #553, opened 2022-09-12
Change default for cdn.flush_manually to true

jacobd91:
Addition to resolution for https://github.com/W3EDGE/w3-total-cache/issues/535 to help prevent unnecessary Cloudfront Invalidation charges

### DIFF
--- a/Cdn_Util.php
+++ b/Cdn_Util.php
@@ -241,7 +241,7 @@ class Cdn_Util {
 	 * @return boolean default value override;
 	 */
 	public static function get_flush_manually_default_override( $cdn_engine = null ) {
-		$override_targets = array( 'S3', 'CF', 'CF2' );
+		$override_targets = array( 's3', 'cf', 'cf2' );
 		return in_array( $cdn_engine, $override_targets, true );
 	}
 }
--- a/Cdn_Util.php
+++ b/Cdn_Util.php
@@ -230,4 +230,18 @@ class Cdn_Util {
 
 		return $file;
 	}
+
+	/**
+	 * Get the override default value for cdn.flush_manually to prevent excessive invalidation charges for S3 CF and CF2.
+	 *
+	 * @since 2.2.6
+	 *
+	 * @param string $cdn_engine CDN engine value.
+	 *
+	 * @return boolean default value override;
+	 */
+	public static function get_flush_manually_default_override( $cdn_engine = null ) {
+		$override_targets = array( 'S3', 'CF', 'CF2' );
+		return in_array( $cdn_engine, $override_targets, true );
+	}
 }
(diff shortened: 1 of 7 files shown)


mavas84 (changes requested):
This should be enabled by default if the following CDN type is selected:
Origin Pull: cf2 - Amazon Cloudfront
Origin Push: cf - Amazon CloudFront Over S3
Origin Push: s3 - Amazon Simple Storage Service (S3)

For all other CDN types it should not be enabled by dafault.

The Only Purge CDN Manually checkbox was added specifically because of this problem in the Performance>CDN>Advanced section. Once checked, any changes are not invalidated automatically, so the cache needs to be purged manually as explained above.
This is not, however, enabled by default.
This checkbox should be enabled by default if one of the following CDN types are selected in the Performance>General Settings>CDN section under CDN type:
Origin Pull: Amazon Cloudfront
Origin push: Amazon Cloudfront over S3
Origin Push: Amazon Simple Storage Service (S3)

mavas84 (approved):

inc/options/cdn.php
```diff
@@ -377,8 +377,11 @@
 
 			<tr>
 				<th colspan="2">
-					<?php $this->checkbox( 'cdn.flush_manually' ); ?>
 					<?php
+					$this->checkbox(
+						'cdn.flush_manually',
+						Cdn_Util::get_flush_manually_default_override( $this->_config->get_string( 'cdn.engine' ) )
```
cssjoe, line 383: This argument disables the checkbox.  A user should be able to toggle the setting regardless of the `cdn.engine` selected.
See https://github.com/W3EDGE/w3-total-cache/blob/5a84ed9f83d7a4fe18c4d59453012b1f444eb4d6/Base_Page_Settings.php#L88

Cdn_Util.php
```diff
@@ -230,4 +230,18 @@ static function replace_folder_placeholders_to_uri( $file ) {
...
+	 * Get the override default value for cdn.flush_manually to prevent excessive invalidation charges for S3 CF and CF2.
+	 *
+	 * @since 2.2.6
+	 *
+	 * @param string $cdn_engine CDN engine value.
+	 *
+	 * @return boolean default value override;
+	 */
+	public static function get_flush_manually_default_override( $cdn_engine = null ) {
+		$override_targets = array( 'S3', 'CF', 'CF2' );
```
cssjoe, line 244: The correct values are: 's3', 'cf', 'cf2'.

cssjoe (changes requested):
If a user selects s3, cf, or cf2, then we should check the `cdn.flush_manually` checkbox and notify the user.
 See the two other comments.

inc/options/cdn.php
jacobd91: Reverted and proper application added to "checkbox" function to override the default boolean for cdn.flush_manually

cssjoe (approved):

PR #577, opened 2022-11-02
Several filesystem related fixes for when WP_HOME is set.

jacobd91:
This is an extension of the fix for 568. I found several other uses of wp_home_url for file-system related items that would likely break if WP_HOME was set. This needs testing under relevant environments for each "fix"

Additionally there are still a decent number of wp_home_url calls in the codebase but as far as I can tell these are either used in URL manipulations or as display values unrelated to file paths


maxicus (commented):
2 points at least, dont care about cdn part much atm. but better to check it manually on those 2 envs

PR #589, opened 2022-11-25
Catch \Exception not \W3TC\Exception

jamieburchell:


cssjoe (approved):
@jamieburchell Thank you for your contribution.  Good catch.

PR #607, opened 2022-12-22
Fix for scheduled posts not flushing PgCache

jacobd91:
This is a compound issue between the original hooks used by Page Cache not applying uniformly between regular pages/posts and custom post types and the Object Cache caching cron jobs and preventing them from executing

This PR changes the hooks so that CPTs are fully supported. A separate PR will address the Object Cache correlated issue


jacobd91:
Issue 605 is effectively the same issue but affecting a different file that this PR doesn't touch

cssjoe (approved):

Issue #606, opened 2022-12-20
E_WARNING: filesize(): stat failed for 


pjcvijay:
W3C Total Cache Version: 2.2.7

I get lots of errors everyday, here is an example of such error:

`E_WARNING: filesize(): stat failed for /var/www/site/public/wp-content/cache/db/options/bb4/f64/bb4f64067d07e62f9bef800340a031b5.php`

This is because following code failed when the file **path** (e.g _bb4/f64/bb4f64067d07e62f9bef800340a031b5.php_) does not exist:

_/wp-content/plugins/w3-total-cache/Cache_File.php (#159)_

`if ( ! $fp || 4 > filesize( $path ) ) {`

Error happens when the pages (URLs) crawled by search engines. 

Thanks!

cssjoe:
The check above (https://github.com/W3EDGE/w3-total-cache/blob/b1441e5be946f0afff207fb0ec45ad46b145bc4e/Cache_File.php#L155) handles checking if the file exists and returns if it does not or the file is not readable.  If the file does exist and is readable but fails to open, then the "fopen" call will return false and cause the condition below it to fail.  If a warning is logged, then something strange may be happening.
If there are errors logged, then check if the file path referenced exists or not in the filesystem.  If it does not then please provide more information on the configuration (plugin configuration and filesystem access type) and how to replicate the issue.  I was not able to replicate the issue.  It appears that it is a database cache file error.  The "dbcache" values from the `wp-content/w3tc-config/master.php` file may help.

PR #611, opened 2022-12-26
Update Cache_File.php

isrgrajan:
added `if(file_exists()) { ... }` with reference to [Issue #606](https://github.com/W3EDGE/w3-total-cache/issues/606)

### DIFF
--- a/Cache_File.php
+++ b/Cache_File.php
@@ -155,9 +155,11 @@ class Cache_File extends Cache_Base {
 		if ( !is_readable( $path ) )
 			return array( null, $has_old_data );
 
-		$fp = @fopen( $path, 'rb' );
-		if ( ! $fp || 4 > filesize( $path ) ) {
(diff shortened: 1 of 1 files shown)


cssjoe:
The check above (https://github.com/W3EDGE/w3-total-cache/blob/b1441e5be946f0afff207fb0ec45ad46b145bc4e/Cache_File.php#L155) handles checking if the file exists and returns if it does not or the file is not readable.  If the file does exist and is readable but fails to open, then the "fopen" call will return false and cause the condition below it to fail.  If a warning is logged, then something strange may be happening.
If there are errors logged, then check if the file path referenced exists or not in the filesystem.  If it does not then please provide more information on the configuration (plugin configuration and filesystem access type) and how to replicate the issue.  I was not able to replicate the issue.  It appears that it is a database cache file error.  The "dbcache" values from the `wp-content/w3tc-config/master.php` file may help.

PR #617, opened 2023-01-04
Fix for object cache not purging options and preventing cron jobs

jacobd91:
This can be replicated by scheduling a CPT with the Object Cache enabled. Without this fix the cron job shouldn't execute. This may also apply to core pages/posts and may be correlated to issue 595


jacobd91:
This PR is correlated to issue 595 related to Page Cache support for scheduled custom post types

cssjoe (approved):

PR #619, opened 2023-01-09
Fix for Util_WpFile.php namespace errors

jacobd91:
The OP indicated that errors resulted during an import but thus far myself and Marko have been unable to replicate said error(s) via multiple import plugins. That being said, the issue is a namespace/availability problem that should be resolved by the proposed tweaks

### DIFF
--- a/Util_WpFile.php
+++ b/Util_WpFile.php
@@ -19,15 +19,18 @@ class Util_WpFile {
 	 * @param string $extra Extra markup for an error message.
 	 */
 	public static function ajax_check_credentials( $extra = null ) {
-		$access_type = get_filesystem_method();
+		// Ensure WP functions are available.
+		require_once ABSPATH . 'wp-load.php';
+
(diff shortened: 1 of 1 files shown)


cssjoe (changes requested):
We should avoid using `wp-load.php`.  If the WordPress global functions are not being recognized, then some additional WordPress files may have to be required.  See https://developer.wordpress.org/reference/functions/wp_filesystem/#more-information and https://developer.wordpress.org/apis/filesystem/ for usage.

>One of the initial challenges for developers using the WP Filesystem API is you cannot initialize it just anywhere. The [request_filesystem_credentials()](https://developer.wordpress.org/reference/functions/request_filesystem_credentials/) function isn’t available until AFTER the wp_loaded action hook, and is only included in the admin area. One of the earliest hooks you can utilize is admin_init.

cssjoe:
I'll close this until we have a reliable way to replicate the issue.

PR #630, opened 2023-01-13
Fix for missing on_comment_status class method.

jacobd91:
To test this simply enable page-cache and navigate to the Comments admin page /wp-admin/edit-comments.php. Then use the quick-links under a given comment to approve/spam/trash the given comment. This should trigger the wp_set_comment_status hook and in turn flush the associated post/feed

### DIFF
--- a/Util_AttachToActions.php
+++ b/Util_AttachToActions.php
@@ -132,6 +132,18 @@ class Util_AttachToActions {
 	}
 
 	/**
+	 * Comment status action
+	 *
+	 * @param integer $comment_id Comment ID.
+	 * @param string  $status Status.
+	 */
+	public function on_comment_status( $comment_id, $status ) {
+		if ( 'approve' === $status || '1' === $status ) {
+			$this->on_comment_change( $comment_id );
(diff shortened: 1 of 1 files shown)


Util_AttachToActions.php
```diff
@@ -131,6 +131,18 @@ public function on_comment_change( $comment_id ) {
 		$this->on_post_change( $post_id );
 	}
 
+	/**
+	 * Comment status action
```
bwmarkle, line 135: What does "Comment status action" mean?

bwmarkle (changes requested):
Please update the PR to include testing instructions.

Util_AttachToActions.php
jacobd91: WordPress features quick-links under comment items on the Comments admin page that allow administrators to alter the status of a comment with a single link. IE approve/spam/trash. These don't appear to trigger the edit hook and instead trigger the wp_set_comment_status hook

cssjoe (approved):

PR #633, opened 2023-01-17
Updated on_comment_status doc block to better describe the hook action

jacobd91:
To test this simply enable page/object/db cache and navigate to the Comments admin page /wp-admin/edit-comments.php. Then use the quick-links under a given comment to approve/spam/trash the given comment. This should trigger the wp_set_comment_status hook and in turn flush the associated post/feed cache(s)


cssjoe (approved):

DbCache_Plugin.php
```diff
@@ -160,7 +160,7 @@ public function on_comment_change( $comment_id ) {
 	}
 
 	/**
-	 * Comment status action
+	 * Comment status action triggered by approve/spam/trash comment quick-links under items on the Comments admin page.
```
bwmarkle, line 163: Is this the only time this method is triggered?

Or is it triggered any time comment status is changed?

bwmarkle (changes requested):

DbCache_Plugin.php
jacobd91: It's trigger anytime the status is changed via the quick-links under the items listed on the Comments admin page

bwmarkle: Correct me if I'm wrong, but I believe your description is inaccurate.

wp_set_comment_status is called more often than just when a quick-link changes the comment:
https://github.com/WordPress/WordPress/search?q=wp_set_comment_status

jacobd91: You are not incorrect. The hook is called more often and by varying actions. I've updated the comment block and included a WP documentation link for the wp_set_comment_status hook

cssjoe (approved):

PR #635, opened 2023-01-18
Release 2.2.12

cssjoe:
To test:
1. Approve a comment while having the page cache enabled.  Also, test with object and database cache enabled.
2. Create or modify a post while having Varnish configured.  The issue occurs when a post is created or modified, and the user has the reverse proxy Varnish setting enabled with Varnish installed and configured.


jacobd91 (approved):
Approved

PR #637, opened 2023-01-23
Fixed evaluations resulting in warnings/errors for non-existent values.

jacobd91:
We were unable to replicate this within the context of a test website, but were able to generate warnings by either manually modifying Util_AttachToActions.php line 78:

`unset($data['post_status']);`

or by CLI testing using:

```
/opt/cpanel/ea-php81/root/bin/php -r '$a = array( "test" => "yes" ); var_dump( $a["nope"] === "yes" );' -d display_errors=on

Warning: Undefined array key "nope" in Command line code on line 1
bool(false)
```

See https://wordpress.org/support/topic/regression-error-on-image-upload-when-memcached-is-enabled/ for reference.

### DIFF
--- a/Util_AttachToActions.php
+++ b/Util_AttachToActions.php
@@ -71,12 +71,12 @@ class Util_AttachToActions {
 		// if attachment changed - parent post has to be flushed
 		// since there are usually attachments content like title
 		// on the page (gallery).
-		if ( 'attachment' === $data['post_type'] ) {
+		if ( isset( $data['post_type'] ) && 'attachment' === $data['post_type'] ) {
 			$post_id = $data['post_parent'];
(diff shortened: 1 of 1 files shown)


jacobd91:
The fix is to check if the value is set for the associative array or object:

```
/opt/cpanel/ea-php81/root/bin/php -r '$a = array( "test" => "yes" ); var_dump( isset($a["nope"]) && $a["nope"] === "yes" );' -d display_errors=on
bool(false)
```

bwmarkle:
@jacobd91 what version of memcached are you running in the test environment?

cssjoe (approved):
The reporting user said that our fix works.

PR #638, opened 2023-01-26
Fix php 8.2 depreciations

8ctopus:
Fixes:
- Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /sites/en/wordpress/web/app/plugins/w3-total-cache/Minify_Environment.php on line 430
- Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /sites/en/wordpress/web/app/plugins/w3-total-cache/Minify_Environment.php on line 495

(line numbers are for latest release 2.2.12)


cssjoe (approved):

cssjoe:
@8ctopus Thank you for your contribution.  I found another instance of the issue and included it.
Reference: https://www.php.net/manual/en/migration82.deprecated.php

PR #643, opened 2023-01-31
Fixes conflict with GuzzleHttp 6.x library in Azure CDN library and a…

### LINKED ISSUE #642: GuzzleHttp library in Azure CDN provider will conflict with newer guzzlehttp/guzzle composer dependencies in other plugins
Aloha, W3 Total Cache includes an older GuzzleHttp library (`v6.x`, I believe) as part of the Microsoft Azure Storage CDN provider:
https://github.com/W3EDGE/w3-total-cache/tree/master/lib/Azure/GuzzleHttp

If W3 Total Cache is configured to use the Azure CDN provider, and if other installed plugins include the `guzzlehttp/guzzle` composer library v7.x or newer, a fatal error will be triggered when W3 Total Cache tries to load the older GuzzleHttp library. For example:
`PHP Fatal error:  Cannot redeclare GuzzleHttp\describe_type() (previously declared in /var/www/html/wp-content/plugins/authorizer/vendor/guzzlehttp/guzzle/src/functions.php:16) in /var/www/html/wp-content/plugins/w3-total-cache/lib/Azure/GuzzleHttp/functions.php on line 41`
https://github.com/guzzle/guzzle/blob/master/src/functions.php#L5-L19

The root cause is the conditional check for the `GuzzleHttp\uri_template` function in `functions_include.php`:
https://github.com/W3EDGE/w3-total-cache/blob/master/lib/Azure/GuzzleHttp/functions_include.php#L4
Guzzle 7.x removed this function, so the conditional passes and tries to load the library, failing once it tries to redeclare the next function, `GuzzleHttp\describe_type`.

This should be a quick fix in W3 Total Cache (unless you want to move the entire dependency into composer, but I assume that could have side effects). Just check `function_exists('GuzzleHttp\describe_type')` instead of `function_exists('GuzzleHttp\uri_template')`.
https://github.com/W3EDGE/w3-total-cache/blob/master/lib/Azure/GuzzleHttp/functions_include.php#L4

For reference, the guzzle folks rolled out a similar fix in 7.0.1:
https://github.com/guzzle/guzzle/pull/2699/files

Thanks!

figureone:
…ny guzzle 7.x library loaded in other plugins

Similar fix is in guzzle 7.0.1: https://github.com/guzzle/guzzle/pull/2699/files

Fixes #642

### DIFF
--- a/lib/Azure/GuzzleHttp/functions_include.php
+++ b/lib/Azure/GuzzleHttp/functions_include.php
@@ -1,6 +1,6 @@
 <?php
 
 // Don't redefine the functions if included multiple times.
-if (!function_exists('GuzzleHttp\uri_template')) {
+if (!function_exists('GuzzleHttp\describe_type')) {
     require __DIR__ . '/functions.php';
 }


cssjoe (approved):

cssjoe:
@figureone Thank you for your contribution.  We'll try out this change.  We do have to look into updating our composer dependency and maybe loading it instead of a piece of an older version.

PR #649, opened 2023-02-14
2.3.0 BrowserCache Environment Rules Fix

jacobd91:
The get_required_rules method features a switch that checks for the appropriate environment rules, ie apache, nginx, lightspeed, etc. This switch was missing a default which resulted in the $rewrite_rules being null if no match was found resulting in an error.

This PR adds a default switch case that will set $rewire_rules to an empty array if no environment match is found.

Reference topic: [php-error-after-upgrading-2-3-0](https://wordpress.org/support/topic/php-error-after-upgraded-2-3-0/)

### DIFF
--- a/BrowserCache_Environment.php
+++ b/BrowserCache_Environment.php
@@ -63,8 +63,9 @@ class BrowserCache_Environment {
 	 * @return array
 	 */
 	public function get_required_rules( $config ) {
-		if ( !$config->get_boolean( 'browsercache.enabled' ) )
-			return null;
+		if ( ! $config->get_boolean( 'browsercache.enabled' ) ) {
+			return array();
+		}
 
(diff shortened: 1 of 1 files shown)


cssjoe (changes requested):
Add `$rewrite_rules = array();` at the beginning of the method and change https://github.com/BoldGrid/w3-total-cache/pull/649/files#diff-b724292a2beb84ace237d5a55ff2df5c83adfbcec30872433c176b79c7cd24bdR67 to `return array();` or `return $rewrite_rules;`.
The method must return an array.

The fix should have been in one commit and formatting in another.  It's a small commit, so it's not a big deal, but for anyone reviewing, it really helps to be able to see only the fix.  Too much going on can let someone miss an important change.

cssjoe (approved):

PR #652, opened 2023-02-16
Update w3tc_modify_query_obj()

### LINKED ISSUE #651: Maintain other meta_query settings in pre_get_posts
My plugin [Videopack ](https://wordpress.org/plugins/video-embed-thumbnail-generator/) modifies the Media Library [meta_query using the pre_get_posts filter](https://github.com/kylegilman/video-embed-thumbnail-generator/blob/e6c604f10bc9dae86e47ab216a59624563b9a6e4/src/admin/videopack-admin.php#L3339) in order to avoid showing attachments created by the plugin. It looks like w3tc does something similar except it overwrites any other meta_query data in the function [w3tc_modify_query_obj ](https://github.com/BoldGrid/w3-total-cache/blob/d2e7c4437799aeb07763d75a768f63d1df303ca5/Root_Loader.php#L238) and disables the meta_query my plugin has added. Could you modify this function to maintain the modified meta_query? I found [this explanation](https://wordpress.stackexchange.com/questions/72099/can-i-exclude-a-post-by-meta-key-using-pre-get-posts-function/72126#72126) helpful.

cssjoe:
To test: Install https://wordpress.org/plugins/video-embed-thumbnail-generator/ to create attachments that the plugin hides in the Media Library.  Activate the W3TC Image Service extension, and optimize/convert an image.  All conversions/copies of image attachments should be hidden.
Fixes #651

Dev notes:
When debugging while having only W3TC activated, the `$query->get( 'meta_query' )` call returned an empty string.  That's why I type-casted it to an array since its use always requires an array.

### DIFF
--- a/Root_Loader.php
+++ b/Root_Loader.php
@@ -235,15 +235,14 @@ class Root_Loader {
 			return;
 		}
 
-		$query->set(
-			'meta_query',
-			array(
-				array(
-					'key'     => 'w3tc_imageservice_file',
-					'compare' => 'NOT EXISTS',
-				),
-			)
+		// Get the existing meta query array, add ours, and then save it.
+		$meta_query   = (array) $query->get( 'meta_query' );
+		$meta_query[] = array(
+			'key'     => 'w3tc_imageservice_file',
+			'compare' => 'NOT EXISTS',
 		);
+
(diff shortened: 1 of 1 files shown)


jacobd91 (approved):
Tested and approved

PR #653, opened 2023-02-16
Fix for sns.php namespace error

jacobd91:
To replicate the error either directly access /wp-content/plugins/w3-total-cache/pub/sns.php in a browser which will result in a fatal error. Reviewing the error log will show
```
[15-Feb-2023 15:58:06 UTC] PHP Fatal error:  Uncaught Error: Class 'Aws\Sns\Message' not found in /home/jacobd8/public_html/wp-content/plugins/w3-total-cache/Enterprise_SnsServer.php:18
Stack trace:
#0 /home/jacobd8/public_html/wp-content/plugins/w3-total-cache/pub/sns.php(58): W3TC\Enterprise_SnsServer->process_message(NULL)
#1 {main}
  thrown in /home/jacobd8/public_html/wp-content/plugins/w3-total-cache/Enterprise_SnsServer.php on line 18
```

The better method is to use a tool like PostMan to submit JSON such as
`{"Type":"Notification","Message":"{\"blog_id\":\"1\",\"host\":\"https://jacobdboldgrid.com\"}"}`

With the fix applied this error will no longer occur, however there may be additional issues with sns.php as in testing other errors occurred but that could be due to unfamiliarity with the processing code and proper input data format

With the above JSON it appeared to process correctly initially but appeared to have a "double run" where the 2nd contained no data, resulting in an error. I was unable to determine the cause


cssjoe (changes requested):
Resolve merge conflicts.  The `vendor` directory has bene removed and is now built int Travis CI.

The classes were separated out of the main package.  See https://github.com/aws/aws-sdk-php/pull/53#issuecomment-117443613
We can add `composer require aws/aws-php-sns-message-validator` to add the package to the `composer.json` file.

cssjoe:
Moved to https://github.com/BoldGrid/w3-total-cache/pull/661

PR #663, opened 2023-03-09
Fix for symlinked cache directory causing folder/file creation to partially fail

jacobd91:
This issue occurs if the cache directory located in wp-content is moved to another location and replaced with a symbolic link. This causes both Minify and PageCache using `file` method to partially fail during the folder/file creation process in that it will only create the first element in the path i.e. `cache/minify/1/287.js` would only create the `cache` directory before failing. This would require refreshing the page or performing an admin action repeatedly until all folders/files are created

Additionally if the minify cache folder is missing it required an admin action to be performed in order for the required minify folder path to be generated which prevented minified files from being served. This appears to only occur if the minify folder fails to generate on admin action or somehow is manually deleted. Performing an admin action resolves this including saving settings or flushing the cache

With this fix applied both single and multisite successfully generate all required folders and files without this partial fail occurring. Additionally this fix will create the required minify folder path via the front end if the folder(s) are missing

### DIFF
--- a/Minify_Plugin.php
+++ b/Minify_Plugin.php
@@ -170,6 +170,7 @@ class Minify_Plugin {
 		$embed_extsrcjs = false;
 		$buffer = apply_filters( 'w3tc_minify_before', $buffer );
 
+		// If the minify cache folder is missing minify fails. This will generate the minify folder path if missing.
 		$minify_environment = Dispatcher::component( 'Minify_Environment' );
 		try {
 			$minify_environment->fix_on_wpadmin_request( $this->_config, true );
--- a/Minify_Plugin.php
+++ b/Minify_Plugin.php
@@ -170,7 +170,10 @@ class Minify_Plugin {
 		$embed_extsrcjs = false;
 		$buffer = apply_filters( 'w3tc_minify_before', $buffer );
 
-
+		$minify_environment = Dispatcher::component( 'Minify_Environment' );
+		try {
+			$minify_environment->fix_on_wpadmin_request( $this->_config, true );
+		} catch ( \Exception $e ) {}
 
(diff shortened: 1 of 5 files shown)


cssjoe (changes requested):
Some PHPUnit tests failed.

`Exception: Can't write to file /master.php`

`UnexpectedValueException: RecursiveDirectoryIterator::__construct(/tmp/wordpress/wp-content/uploads): failed to open dir: No such file or directory`

cssjoe:
@jacobd91 I made some small changes.  Please test again.

Minify_Plugin.php
```diff
@@ -170,7 +170,11 @@ function ob_callback( $buffer ) {
 		$embed_extsrcjs = false;
 		$buffer = apply_filters( 'w3tc_minify_before', $buffer );
 
-
+		// If the minify cache folder is missing minify fails. This will generate the minify folder path if missing.
+		$minify_environment = Dispatcher::component( 'Minify_Environment' );
```
bwmarkle, line 174:
```
/**
 * If the minify cache folder is missing, create it.
 */
function maybe_create_minify() {
  $minify_environment = Dispatcher::component( 'Minify_Environment' );
  
  try {
     $minify_environment->fix_on_wpadmin_request( $this->_config, true );
  } catch ( \Exception $e ) {}
}



// If the minify cache folder is missing minify fails. This will generate the minify folder path if missing.
$util->maybe_create_minify();


PHPUNIT TEST LOGIC
1. If it doesn't exist, create the minify folder.
2. Assert that it exists
3. Delete it
4. Assert that it does not exist
5. $util->maybe_create_minify();
6. Assert that it exists
```

cssjoe (approved):

PR #669, opened 2023-03-28
Fix Memcached test using SASL auth

### LINKED ISSUE #448: Memcached test using SaslAuth
From https://wordpress.org/support/topic/memcached-tests-when-use-sasl/

Referenced:
https://github.com/W3EDGE/w3-total-cache/blob/master/Generic_AdminActions_Test.php#L160
https://github.com/W3EDGE/w3-total-cache/blob/master/Util_Installed.php#L128

The credentials are added in the Cache_Memcached class when used, if the passed config includes them:
https://github.com/W3EDGE/w3-total-cache/blob/master/Cache_Memcached.php#L83-L87

The test retrieves the Cache_Memcached instance, which should use the credentials, if passed:
https://github.com/W3EDGE/w3-total-cache/blob/master/Cache.php#L51

The test and check if installed should both pass the username and password from the configuration, if set.

paddyohanlon:
Fixes #448.

When testing a Memcached cache, in addition to passing the configured servers, also pass binary protocol, username and password. This allows the test to be run correctly when SASL authentication is used (when a username and password are supplied). The binary protocol option is also passed as SASL auth with Memcached requires the binary protocol to be used.

### DIFF
--- a/DbCache_Plugin_Admin.php
+++ b/DbCache_Plugin_Admin.php
@@ -52,8 +52,11 @@ class DbCache_Plugin_Admin {
 
 		if ( $c->get_string( 'dbcache.engine' ) == 'memcached' ) {
 			$memcached_servers = $c->get_array( 'dbcache.memcached.servers' );
+			$memcached_binary_protocol = $c->get_boolean( 'dbcache.memcached.binary_protocol' );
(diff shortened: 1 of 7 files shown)


cssjoe (approved):
@paddyohanlon Thank you for your contribution.  These changes will most likely be included in our next release.

PR #670, opened 2023-03-30
Update Util_WpFile.php

isrgrajan:
Fix for  PHP Fatal error: Uncaught Error: Call to undefined function W3TC\get_filesystem_method()


cssjoe (changes requested):
Thank you for your proposed fix.  The new code should be placed in the `ajax_check_credentials` method before the `get_filesystem_method()` call.

avonville:
updated PR in #978 moved conditional in `ajax_check_credentials()`. review and merge will be done in #978. Closing this PR.

PR #677, opened 2023-04-28
fix for nginx header|permissions fields being a string instead of array

dcordell99:
In the Browser Cache settings for Nginx for the Header/Permission Policies, any input in any field (none, src, etc.) will cause a critical error on the site when you save. It appears this happens because the $rules variable gets converted from an Array to a String. To fix this issue, I created a variable to house the concatenated string with the correct nginx "add_header Feature-Policy" and "add_header Permission-Policy" string prepend. Then pushed into the $rules array. This sets the nginx.conf correctly and allows the nginx server to pass testing / restart.

### DIFF
--- a/BrowserCache_Environment_Nginx.php
+++ b/BrowserCache_Environment_Nginx.php
@@ -344,11 +344,13 @@ class BrowserCache_Environment_Nginx {
 				}
 
 				if ( ! empty( $feature_v ) ) {
-					$rules .= '    Header set Feature-Policy "' . implode( ';', $feature_v ) . "\"\n";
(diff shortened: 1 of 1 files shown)


cssjoe:
This pull request is outdated and now has merge conflicts.  If there is still an issue in or after version 2.6.1 then please let us know.

dcordell99:
> This pull request is outdated and now has merge conflicts. If there is still an issue in or after version 2.6.1 then please let us know.

Hi Joe. The issue still persists on 2.6.1. The same changes in this request also still fixes the issue.

cssjoe:
@dcordell99 The issue has been fixed on https://github.com/BoldGrid/w3-total-cache/pull/779 and another pull request.  It will be included in the next release.  Thank you for letting us know.

PR #680, opened 2023-05-03
Set permissions on created cache files to value of FS_CHMOD_FILE

andyexeter:
This was already been done in other parts of the codebase e.g. https://github.com/BoldGrid/w3-total-cache/blob/3a094493064ea60d727b3389dee813639860ef49/Util_File.php#L398-L401

So I have taken the same code and applied it in other places files are created.

This is needed for example so a web server user and CLI user in the same group both have permission to clear cache files when `FS_CHMOD_FILE` is set to `0664`.

### DIFF
--- a/Cache_File_Generic.php
+++ b/Cache_File_Generic.php
@@ -60,6 +60,12 @@ class Cache_File_Generic extends Cache_File {
 		@fputs( $fp, $var['content'] );
 		@fclose( $fp );
 
+		$chmod = 0644;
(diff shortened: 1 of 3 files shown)


andyexeter:
Are any maintainers able to take a look at this?

cssjoe:
We'll test this.  The new code needs to be updated for PHPCS/WPCS.

avonville:
Closing this PR, new pr with fixes and coding standard updates #977

PR #681, opened 2023-05-03
Fix GA switch for development

cssjoe:
This is to avoid the W3TC_DEBUG constant from switching Google Analytics to the development property.  I added a new constant check for "W3TC_DEVELOPER", which will only change the GA property ID to development.


jacobd91 (approved):
Short and sweat. Approved

PR #707, opened 2023-06-27
Enhancement 583

jacobd91:
This PR implements the wp_get/add/set/delete_multiple methods utilized by memcached, redis, etc. In testing with W3TC object cache configured to use either memcached or redis it appears to be utilizing the get/add/delete multi methods, however it did not appear to utilize the set multi method in testing. This may not be typically used by memcached or redis, and may be intended for more direct use rather than general cache functionality (speculation)


spacedmonkey:
@cssjoe This doesn't really have an performance benefit. You really want to use either https://www.php.net/manual/en/memcached.getmulti.php or https://redis.io/commands/mget/ to get multiple keys in a single request.

PR #708, opened 2023-06-30
dbcluster fix

maxicus:


mavas84:
I've tested the PR, with the fixes provided and now the DB cluster works as expected without problems with page rendering.
``

cssjoe (approved):

PR #711, opened 2023-07-07
Fix Travis CI

cssjoe:
I updated the WordPress test version to 6.2 and updated the installer script using https://github.com/wp-cli/sample-plugin/blob/master/bin/install-wp-tests.sh
There is no test for this PR; just review the Travis CI build results.


jacobd91 (approved):
Reviewed and approved

PR #715, opened 2023-07-18
Fix php 8.2 deprecation

8ctopus:
Fix php 8.2 creation of dynamic properties.
Please check the variable type for `$_page_group` as it was a string in my case, but it could be more vast.


cssjoe:
@8ctopus Thank you for your contribution.  `$_page_group` is a string.  The doc block comments can use some updating.  We'll handle that at a later date.  We'll get these changes included in our 2.4.0 release this week.

PR #719, opened 2023-08-02
Validate if file exists before delete

jaimearroyonavia:
I'am using the Elastic APM and we see this unlink without checking if the file exist, and is generating a lot of E_WARNING log


cssjoe:
The `@` should silence errors.  There may be a way to ignore them if coming through on APM.  An extra file existence check would double the filesystem operations at that point.  We haven't seen any errors logged at that point; since it has the `@`.

PR #724, opened 2023-08-09
add a fallback for older version of wp-content/db.php

maxicus:
which doesn't pass config parameter


jacobd91 (commented):
Along the lines of what I was thinking. Beat me to it and approved

jacobd91 (approved):

PR #727, opened 2023-08-16
ability to activate fragmentcache

maxicus:
engine selection has to be allowed when extension is not frontend-active. backend-activity should be enough for UI


cssjoe (approved):

jacobd91 (commented):
Approved

jacobd91 (approved):

PR #743, opened 2023-09-21
Fix Nginx rules and php 8.2 deprecated error

cssjoe:
Fixed https://github.com/BoldGrid/w3-total-cache/blob/2.4.1/BrowserCache_Environment_Nginx.php#L346-L352 where lines should be added into the rules array and not concatenating a string.

Fixed incorrect class property name at https://github.com/BoldGrid/w3-total-cache/blob/2.4.1/Mobile_Base.php#L13 to what is used at https://github.com/BoldGrid/w3-total-cache/blob/2.4.1/Mobile_Base.php#L23


jacobd91 (approved):
Approved

PR #753, opened 2023-10-05
Preload Requests pro feature

jacobd91:
The feature is implemented similarly to the Delay Scripts feature in that it's an internal extension of the User Experience

The "enable" checkbox appears under the User Experience block of the General Settings page. Once enabled it should contain a link to the advanced settings or you can find it under the User Experience advanced settings page. It will contain it's own settings metabox featuring textareas for dns-prefetch, preconnect, preload-css/js/fonts/images/videos/audio/documents

Add domains to dns-prefetch and/or preconnect, one per line.

Add full URLs for css/js/fonts/images/videos/audio/documents that should be preloaded, one per line.

To test configure the feature and then flush all. Then inspect the DOM for the <link> headers for each item. Then test both before and after with Google PageSpeed or GTMetrix to see results (My testing had limited impact due to site simplicity so more complex sites should be more telling)

### DIFF
--- a/UserExperience_Preload_Requests_Extension.php
+++ b/UserExperience_Preload_Requests_Extension.php
@@ -2,7 +2,7 @@
 /**
  * File: UserExperience_Preload_Requests_Extension.php
  *
- * Controls the defer JS feature.
+ * Controls the Preload Requests feature.
  *
  * @since 2.5.1
  *
@@ -61,7 +61,7 @@ class UserExperience_Preload_Requests_Extension {
 	}
 
 	/**
-	 * Renders the user experience defer JS settings page.
+	 * Renders the user experience Preload Requests settings page.
 	 *
 	 * @since 2.5.1
 	 *
--- a/UserExperience_Preload_Requests_Extension.php
+++ b/UserExperience_Preload_Requests_Extension.php
@@ -183,6 +183,19 @@ class UserExperience_Preload_Requests_Extension {
 			echo '<link rel="preload" href="' . esc_url( $url ) . '" as="document">';
 		}
 	}
+
+	/**
+	 * Gets the enabled status of the extension.
+	 *
+	 * @since 2.5.1
+	 *
+	 * @return bool
+	 */
+	public static function is_enabled() {
+		$config            = Dispatcher::config();
+		$extensions_active = $config->get_array( 'extensions.active' );
+		return Util_Environment::is_w3tc_pro( $config ) && array_key_exists( 'user-experience-preload-requests', $extensions_active );
+	}
 }
 
 $o = new UserExperience_Preload_Requests_Extension();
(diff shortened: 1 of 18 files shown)


UserExperience_Preload_Requests_Extension.php
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	/**
+	 * Applies the Preload Requests headers for wp_head and admin_head.
+	 *
+	 * @since 2.5.1
+	 *
+	 * @return void
+	 */
+	public function w3tc_preload_requests_headers() {
```
cssjoe, line 140: All of the [esc_url()](https://developer.wordpress.org/reference/functions/esc_url/#more-information) calls will prepend "http://" to all URLs without a schema.  I used the suggested format "//fonts.googleapis.com" and I got mixed content errors with browsing an HTTPS page.

cssjoe (changes requested):

UserExperience_Preload_Requests_Page_View.php
```diff
@@ -0,0 +1,101 @@
...
+<table class="form-table">
+	<?php
+	Util_Ui::config_item(
+		array(
+			'key'         => array( 'user-experience-preload-requests', 'dns-prefetch' ),
+			'label'       => esc_html__( 'DNS Prefetch Domains:', 'w3-total-cache' ),
+			'control'     => 'textarea',
+			'description' => esc_html__( 'Specify domains whose DNS should be prefetched by browsers via the "dns-prefetch" header. Domains entries with no protocol (//somedomain.com) will default to non-SSL HTTP (http://somedomain). Include one domain entry per line, e.g. (https://cdn.domain.com, https://fonts.googleapis.com, https://www.google-ananlytics.com, etc.)', 'w3-total-cache' ),
```
cssjoe, line 29: Since it applies to all, the "no protocol" part should go in the main description instead of the separate items.

cssjoe (changes requested):
The descriptions need more work.  The main description is long.  The mention of "no protocol" should be in the main description and not in a couple of the setting descriptions when it applies to all of them.  If they all require full URLs, then use "Include one full URL address per line." for each of them.  "non-SSL" refers to the old SSL protocol; TLS is the new type, but it should most likely say "non-HTTPS".

UserExperience_Preload_Requests_Page_View.php
jacobd91: Descriptions and instructions updated

UserExperience_Preload_Requests_Extension.php
jacobd91: Due to the fact that the protocol would be unknown for entries omitting it, there would be no way to determine the correct protocol. As a result I left the esc_url as is and updated the descriptions to indicate that values omitting the protocol would default to non-HTTPS

cssjoe (approved):

PR #761, opened 2023-10-27
Update UserExperience_LazyLoad_Mutator.php

garymarkfuller:
Check if $url_match[2] is set.


cssjoe:
@garymarkfuller Thank you for your contribution.  A similar fix has been included in our pending 2.7.0 release, which will be deployed soon.

PR #773, opened 2023-11-16
Fix Bunny CDN library upload

cssjoe:
To test, enable Bunny CDN for static assets.  You do not have to authorize the CDN to replicate the issue.  Before the update, try to upload something to the WordPress Media Library with Bunny CDN enabled.  An error message will be raised.  After applying the fix, the error will not be raised and the file will be uploaded without issue.


jacobd91 (approved):
Simple fix. Approved

PR #774, opened 2023-11-16
Allow Bunny CDN custom hostname changes

### LINKED ISSUE #768: Bunny CDN custom hostname
At the moment, if the custom hostname is created in the BunnyCDN, it cannot be used as an option in the W3 Total Cache when BunnyCDN is enabled as a static CDN feature.
Add the ability to manually add a custom hostname that will be used for rewriting the CDN URLs
Reference topic: https://wordpress.org/support/topic/bunny-net-custom-hostname-support/

cssjoe:
Adds the ability to change the Bunny CDN custom hostname used when rewriting asset URLs on pages.  It does not change anything at Bunny.Net.
See https://github.com/BoldGrid/w3-total-cache/issues/768
Fixes #768

### DIFF
--- a/Cdn_BunnyCdn_Page_View.js
+++ b/Cdn_BunnyCdn_Page_View.js
@@ -80,14 +80,21 @@ jQuery(function($) {
 		.on('change', '#w3tc-origin-url', function() {
 			var $this = $(this);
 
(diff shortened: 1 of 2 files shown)


jacobd91 (approved):
Approved

PR #775, opened 2023-11-20
Fixes #770

### LINKED ISSUE #770: Cloudflare extension options not updating
In Cloudflare extensions, some options are not updating when changed from the dropdown.
CloudFlare: SSL
By default the option selected in the CloudFlare: SSL dropdown is Full (https everywhere). If the option "Strict" is selected from the dropdown and the Cloudflare settings are saved within the extension, the configuration is not updated and we the error appears:
Failed to update CloudFlare settings:
Failed to update option ssl: Invalid value for zone setting SSL
image
If the option is changed directly in the Cloudflare control panel, SSL/TLS>Overview to Full (Strict), the configuration is updated and the dropdown menu in the W3TC Cloudflare extension is blank:
image

This is also the case with the Browser TTL settings. If this is changed in the W3 Total Cache to anything other than the default, for example, 31536000s to match the W3TC BC settings, upon the saving of the Cloudflare settings the error is thrown:
Failed to update CloudFlare settings:
Failed to update option browser_cache_ttl: Invalid value for zone setting browser_cache_ttl
image

cssjoe:
See https://github.com/BoldGrid/w3-total-cache/issues/770

The numeric values should be passed to the Cloudflare API as integers.
The Max Uploads value must be one of the accepted values.

### DIFF
--- a/Extension_CloudFlare_Api.php
+++ b/Extension_CloudFlare_Api.php
@@ -111,6 +111,11 @@ class Extension_CloudFlare_Api {
 
 
 	public function zone_setting_set( $name, $value ) {
+		// Convert numeric values to the integer type.
+		if ( is_numeric( $value ) ) {
+			$value = intval( $value );
+		}
+
 		return $this->_wp_remote_request( 'PATCH',
 			self::$_root_uri . '/zones/' . $this->_zone_id . '/settings/' . $name,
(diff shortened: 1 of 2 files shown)


jacobd91 (approved):
Approved

PR #776, opened 2023-11-21
Fix lazy loading when using delay scripts

cssjoe:


mavas84 (commented):
The proposed changes are working and Delay scripts feature does not prevent Lazyloading of images with these changes

jacobd91 (approved):
Approved

PR #779, opened 2023-11-28
Fix Nginx feature and permission policy header syntax

cssjoe:
See https://wordpress.org/support/topic/feature-policy-permissions-policy-issue-with-nginx/

To test, set the `Feature-Policy` and `Permissions-Policy` headers in the Browser Cache Security Headers section (`wp-admin/admin.php?page=w3tc_browsercache`) when using Nginx as a web server.  Review the generated "nginx.conf" file in the WordPress ABSPATH directory.


jacobd91 (approved):
Appoved

PR #781, opened 2023-12-06
Fixes an uncaught type error on the Install page.

nitz:
When `Util_Rule::can_check_rules()` returns `false` (as it does on the Caddy web server,) `$other_areas` is left undefined which causes an uncaught type error on https://github.com/BoldGrid/w3-total-cache/blob/f043f6eb1d4c3a8327b8c63867a4d37296c94dff/inc/options/install.php#L281

This change has `$other_areas` behave the same as `$rewrite_rules_descriptors`.


cssjoe (approved):

cssjoe:
@nitz Thank you for your contribution.

Issue #793, opened 2024-01-03
PHP 8 deprecated errors


cssjoe:
PHP 8.1:

"wp-admin/upload.php" -- Change media when an Amazon Cloudfront CDN is configured.
```
PHP Deprecated:  Aws\Credentials\Credentials implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary) in wp-content/plugins/w3-total-cache/vendor/aws/aws-sdk-php/src/Credentials/Credentials.php on line 8
PHP Deprecated:  Return type of Aws\HandlerList::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in wp-content/plugins/w3-total-cache/vendor/aws/aws-sdk-php/src/HandlerList.php on line 308
PHP Deprecated:  Return type of Aws\Api\AbstractModel::offsetExists($offset) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in wp-content/plugins/w3-total-cache/vendor/aws/aws-sdk-php/src/Api/AbstractModel.php on line 41
PHP Deprecated:  Return type of Aws\Api\AbstractModel::offsetGet($offset) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in wp-content/plugins/w3-total-cache/vendor/aws/aws-sdk-php/src/Api/AbstractModel.php on line 30
PHP Deprecated:  Return type of Aws\Api\AbstractModel::offsetSet($offset, $value) should either be compatible with ArrayAccess::offsetSet(mixed $offset, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in wp-content/plugins/w3-total-cache/vendor/aws/aws-sdk-php/src/Api/AbstractModel.php on line 36
PHP Deprecated:  Return type of Aws\Api\AbstractModel::offsetUnset($offset) should either be compatible with ArrayAccess::offsetUnset(mixed $offset): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in wp-content/plugins/w3-total-cache/vendor/aws/aws-sdk-php/src/Api/AbstractModel.php on line 46
[...]

PR #797, opened 2024-01-11
#793 Fix deprecation errors dynamic property assignment

michaelw85:
Fixes a couple `Creation of dynamic property is deprecated` errors.

Question: Should the libs be fully updated instead of locally modified? If so, they should be moved to composer in my opinion. 
To keep impact low I chose to only fix the deprecation warnings for now.


cssjoe (approved):
@michaelw85 Thank you for your contribution.  We will be reviewing the libraries in the future.  This pull request will be included in our next release.

PR #805, opened 2024-01-23
Upgrade/Purchase Modal Update

jacobd91:
This PR simply moves the "Go Faster With Pro" button in the upgrade modal to the bottom and changes the appearance and text to "Subscribe to Go Faster Now". It also tweaks the purchase modal so that the "close window" matches that of the upgrade modal as well as making the sizing of both consistent for width


jacobd91:
Brad will be reviewing the changes possibly locally so there may be further tweaks

cssjoe (dismissed):

cssjoe (approved):

PR #820, opened 2024-02-08
Fix flush all target error. Fix browsercache replace target error.

jacobd91:


cssjoe:
Fixing on the release branch.

PR #849, opened 2024-03-27
Remove CSS/JS Expansion

jacobd91:
This PR fixes an issue with single rules incorrectly saving and display the defined file path on page reload

Mainly however, it expands the single rule to now include a radio button for "Exclude" and "Include". 

Exclude will only remove the defined file on URLs that match the values defined in the textarea

Include will remove the defined file on all URLs except those that match the values defined in the textarea

To test you must have PRO and navigate to **General Settings => User Experience** and enable the **Remove Unwanted/Unused CSS/JS**. Then navigate to the User Experience advanced settings page and scroll to the **Remove CSS/JS Individually** section. Then click **Add** and define a rule to match a given CSS/JS file you wish to control. Follow the above for exclude/include and define the URLs you wish for the rule to apply to. Then verify via the browser (may require flush depending on config)


UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -11,14 +11,14 @@ jQuery(function() {
...
+			var single_path = prompt('Enter CSS/JS URL.');
 
-			if (single !== null) {
+			if (single_path !== null) {
 				var exists = false;
 
 				jQuery('.remove_cssjs_singles_path').each(
 					function() {
-						if (jQuery(this).html() == single) {
+						if (jQuery(this).html() == single_path) {
```
cssjoe, line 21: >Strict equality checks (===) must be used in favor of abstract equality checks (==).

https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#equality

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -27,21 +27,32 @@ jQuery(function() {
 				);
 
 				if (!exists) {
+					var single_id = single_path.replace(/[^\w-]/g, '_');
```
cssjoe, line 30: >Variable and function names should be full words, using camel case with a lowercase first letter.

https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#naming-conventions

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -27,21 +27,32 @@ jQuery(function() {
...
 						'</td>' +
 						'</tr>' +
 						'<tr>' +
-						'<th><label for="remove_cssjs_singles_' + single + '_includes">Remove on these pages:</label></th>' +
+						'<th><label class="remove_cssjs_singles_' + single_id + '_includes_label" for="remove_cssjs_singles_' + single_id + '_includes">Exclude on these pages:</label></th>' +
 						'<td>' +
-						'<textarea id="remove_cssjs_singles_' + single + '_includes" name="user-experience-remove-cssjs-singles[' + single + '][includes]" rows="5" cols="50"></textarea>' +
-						'<p class="description">Specify relative/absolute page URLs that the above CSS/JS should be removed from. Include one entry per line.</p>' +
+						'<textarea id="remove_cssjs_singles_' + single_id + '_includes" name="user-experience-remove-cssjs-singles[' + single_path + '][includes]" rows="5" cols="50" ></textarea>' +
+						'<p class="description remove_cssjs_singles_' + single_id + '_includes_description">Specify relative/absolute page URLs that the above CSS/JS file should be Excluded from. Include one entry per line.</p>' +
```
cssjoe, line 55: All text should be run through translation functions.  You can do this in PHP and use https://developer.wordpress.org/reference/functions/wp_localize_script/ to pass an object to JavaScript.
Also, do not capitalize words like "Exclude" -- apply `<em>` or `<strong>` instead.  Change the wording "be excluded from" to "to be excluded".

cssjoe (changes requested):
When I click "Add" on `wp-admin/admin.php?page=w3tc_userexperience#remove-cssjs-singles` and then click "Cancel" in the popup modal, I get an error in another popup saying "Empty CSS/JS URL!".

When I delete an entry and save the settings, the page reloads and the item remains.

Review https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/ and apply standards to the JS file.

UserExperience_Remove_CssJs_Page_View.js
jacobd91: Fixed

UserExperience_Remove_CssJs_Page_View.js
jacobd91: Updated all variable and methods to use camel case

UserExperience_Remove_CssJs_Extension.php
```diff
@@ -55,7 +55,7 @@ public function run() {
 
 		Util_Bus::add_ob_callback( 'removecssjs', array( $this, 'ob_callback' ) );
 
-		add_filter( 'w3tc_save_options', array( $this, 'w3tc_save_options' ) );
+		add_action( 'w3tc_config_ui_save-w3tc_userexperience', array( $this, 'w3tc_config_ui_save_w3tc_userexperience' ), 10, 2 );
```
cssjoe, line 58: The callback function name is long.  We try to keep names limited to 3 parts when possible.
  It also messes up the history of the method in Git.

UserExperience_Remove_CssJs_Extension.php
```diff
@@ -195,32 +195,34 @@ public function w3tc_config_key_descriptor( $descriptor, $key ) {
...
-		$new_cssjs_includes = $new_config->get_array( array( 'user-experience-remove-cssjs', 'includes' ) );
-		$new_cssjs_singles  = $new_config->get_array( 'user-experience-remove-cssjs-singles' );
+	public static function w3tc_config_ui_save_w3tc_userexperience( $new_config, $old_config ) {
+		$raw_cssjs_singles = Util_Request::get_array( 'user-experience-remove-cssjs-singles' );
 
-		foreach ( $new_cssjs_singles as $url => $pages ) {
-			if ( is_string( $pages['includes'] ) ) {
+		$new_cssjs_singles = array();
+		foreach ( $raw_cssjs_singles as $url => $pages ) {
+			if ( esc_url( $url ) && is_string( $pages['includes'] ) && ! empty( $pages['action'] ) ) {
```
cssjoe, line 210: It looks like the `esc_url()` call is in the wrong place.  It's not escaping the data before it's used as an array key below.  I know that it was previously implemented in the settings that way, but it's not a good idea to use a URL as an array key.

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -11,14 +11,19 @@ jQuery(function() {
 		'click',
 		'#w3tc_remove_cssjs_singles_add',
 		function() {
-			var single_path = prompt('Enter CSS/JS URL.');
+			var singlePath = prompt('Enter CSS/JS URL.');
```
cssjoe, line 14: `const` is now recommended for this use instead of `var`.  See https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#declaring-variables-with-const-and-let

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -11,14 +11,19 @@ jQuery(function() {
...
 
-			if (single_path !== null) {
+			if (singlePath === null) {
```
cssjoe, line 16: Use Yoda conditions and spacing -- `if ( null === singlePath ) {`.

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -11,14 +11,19 @@ jQuery(function() {
...
+				return;
+			}
+
+			singlePath = singlePath.trim();
+			if (singlePath !== '') {
```
cssjoe, line 21: You can use `if ( singlePath ) {` instead of checking it it's not an empty string.

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -27,41 +32,41 @@ jQuery(function() {
...
 						'<input type="button" class="button remove_cssjs_singles_delete" value="Delete"/>' +
 						'</td>' +
 						'</tr>' +
 						'<tr>' +
-						'<th><label for="remove_cssjs_singles_' + single_id + '_action">Behavior:</label></th>' +
+						'<th><label for="remove_cssjs_singles_' + singleID + '_action">Behavior:</label></th>' +
 						'<td>' +
-						'<label class="remove_cssjs_singles_behavior"><input class="remove_cssjs_singles_behavior_radio" type="radio" name="user-experience-remove-cssjs-singles[' + single_path + '][action]" value="exclude" checked>Exclude</label>' +
-						'<label class="remove_cssjs_singles_behavior"><input class="remove_cssjs_singles_behavior_radio" type="radio" name="user-experience-remove-cssjs-singles[' + single_path + '][action]" value="include">Include</label>' +
+						'<label class="remove_cssjs_singles_behavior"><input class="remove_cssjs_singles_behavior_radio" type="radio" name="user-experience-remove-cssjs-singles[' + singlePath + '][action]" value="exclude" checked>Exclude</label>' +
```
cssjoe, line 50: The text displayed to a user should be translatable.  Use https://developer.wordpress.org/reference/functions/wp_localize_script/ to pass translated text in an object for use in this JavaScript file. 
There are multiple instances of text in this file that should be run through translation functions.

cssjoe (changes requested):
See comments.; address coding standards and localization for translatable text in JS.

UserExperience_Remove_CssJs_Extension.php
jacobd91: I pulled the esc_url and applied it to a variable so it's escaped in use. Using that value as a key was left as is, as changing it wouldn't be compatible with existing users and would require a restructure of all relevant code for the change in data structure storage

UserExperience_Remove_CssJs_Page_View.js
jacobd91: All vars have been changed to const

UserExperience_Remove_CssJs_Page_View.js
jacobd91: Yoda applied

UserExperience_Remove_CssJs_Page_View.js
jacobd91: Logic simplified

UserExperience_Remove_CssJs_Page_View.js
jacobd91: Translations added

cssjoe (changes requested):
The settings page nav link `wp-admin/admin.php?page=w3tc_userexperience#remove-cssjs-singles` says "Remove CSS/JS - Singles", but the section heading says "Remove CSS/JS Individually" -- be consistent in naming.  Maybe use the section heading for the nav link and URL fragment.

Using the old data format, I get a warning when browsing the frontend.
```PHP Warning:  Undefined array key "action" in wp-content/plugins/w3-total-cache/UserExperience_Remove_CssJs_Mutator.php on line 91```
My data included:
```
array (
  0 =>
  array (
    'includes' =>
    array (
      0 => '/',
    ),
  ),
)```

The new functionality works after saving with the new format.  The old format did not exclude my JS.

The "Exclude on these pages:" textarea says "Specify the absolute or relative URLs".  If I add "/", it works for the homepage, but not other pages.  I would assume that it worked the same way as the URL pattern.  Maybe the method of matching is different.

Also, see comments.

Generic_Plugin_Admin.php
```diff
@@ -626,15 +626,32 @@ public function admin_print_scripts() {
...
 			case 'w3tc_userexperience':
 				if ( UserExperience_Remove_CssJs_Extension::is_enabled() ) {
-					wp_enqueue_script(
+					wp_register_script( 'w3tc_remove_cssjs', plugins_url( 'UserExperience_Remove_CssJs_Page_View.js', W3TC_FILE ), array( 'jquery' ), W3TC_VERSION, true );
+
+					wp_localize_script(
 						'w3tc_remove_cssjs',
-						plugins_url( 'UserExperience_Remove_CssJs_Page_View.js', W3TC_FILE ),
+						'W3TCRemoveCssJsData',
 						array(
```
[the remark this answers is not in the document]
jacobd91: Updated

UserExperience_Remove_CssJs_Mutator.php
```diff
@@ -141,24 +155,27 @@ private function is_content_included( $content ) {
...
-						$pages['includes']
+						$config['includes']
 					)
-				)
-			) {
-				return true;
+				);
+
+				// If set to exclude, remove the file if the page matches defined URLs.
+				// If set to include, Remove the file if the page doesn't match defined URLs.
```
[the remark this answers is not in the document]
jacobd91: Updated

UserExperience_Remove_CssJs_Page_View.js
```diff
@@ -4,80 +4,160 @@
...
-	}
+function setRemoveCssjsSinglesPathValues() {
+    jQuery('.remove_cssjs_singles_path').each(
+		function() {
+        	var $inputField = jQuery(this);
+        	var originalValue = $inputField.val();
+        	$inputField.data('originalValue', originalValue);
+    	}
+	);
 }
```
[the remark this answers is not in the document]
jacobd91: Newline added

UserExperience_Remove_CssJs_Extension.php
```diff
@@ -198,10 +198,11 @@ public function w3tc_config_key_descriptor( $descriptor, $key ) {
 	 * @since 2.7.0
 	 *
 	 * @param array $data Array of save data.
+	 * @param array $page String page value.
 	 *
 	 * @return array
 	 */
-	public function w3tc_save_options( $data ) {
+	public function w3tc_save_options( $data, $page ) {
```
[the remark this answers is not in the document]
jacobd91: Condition added to check for page

UserExperience_Remove_CssJs_Mutator.php
```diff
@@ -141,24 +155,27 @@ private function is_content_included( $content ) {
 		);
 
 		// Only removes matched CSS/JS on matching pages.
-		foreach ( $this->singles_includes as $include => $pages ) {
-			if (
-				! empty( $pages )
-				// Check if the given single CSS/JS remove rule URL is present in HTML content.
-				&& strpos( $content, $include ) !== false
+		foreach ( $this->singles_includes as $id => $config ) {
```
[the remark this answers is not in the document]
jacobd91: Updated to $data

UserExperience_Remove_CssJs_Extension.php
```diff
@@ -198,10 +198,11 @@ public function w3tc_config_key_descriptor( $descriptor, $key ) {
...
 		$new_config = $data['new_config'];
```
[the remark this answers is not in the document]
jacobd91: Updated

UserExperience_Remove_CssJs_Extension.php
```diff
@@ -210,17 +211,24 @@ public function w3tc_save_options( $data ) {
...
+					$new_cssjs_singles[ $single_id ]['includes']    = Util_Environment::textarea_to_array( $single_config['includes'] );
+				}
 			}
-		}
 
-		$new_config->set( 'user-experience-remove-cssjs-singles', $new_cssjs_singles );
+			$new_config->set( 'user-experience-remove-cssjs-singles', $new_cssjs_singles );
+		}
 
 		if ( $new_cssjs_includes !== $old_cssjs_includes || $new_cssjs_singles !== $old_cssjs_singles ) {
```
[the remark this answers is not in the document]
jacobd91: Updated

cssjoe (approved):

PR #871, opened 2024-04-26
us-east-1 (default region) is not used in the Cloudfront hostname

cssjoe:
Fixes https://github.com/BoldGrid/w3-total-cache/issues/845

This enhancement also adds a display span for the effective hostname, visible on the settings page.

To test:
- Create or have an existing Cloudfront distribution using an S3 bucket for storage in the "us-east-1" default region.
- Enter the CDN details in the CDN Cloudfront S3 push configuration.
- Use the test button to see that it passes for both US East selections for "bucket".


jacobd91 (approved):
Approved

PR #875, opened 2024-04-26
Add newsletter signup to footer

elanasparkle:
Provides the Newsletter to the footer of the W3 Total Cache plugin. 

image


elanasparkle:
It wont let me add anything but the ticket is here: 

https://github.com/BoldGrid/w3-total-cache/issues/874

bwmarkle:
General feedback:

image

image

image

----

Thoughts on doing something like this?
image

elanasparkle:
Updated everything should look like this now. 

image

cssjoe (changes requested):
After submitting an email address, the text below is in a small area and pushes the footer up encroaching on the bottom of the page body.  The same thing happens if you click the field and then click elsewhere; the error message increases the height of the footer.

Also, there should be a newline at the end of the file.

jacobd91 (dismissed):
The subscription appears to work and the UI for it is mostly okay. Joe's concern about the error/success message altering the height of the footer isn't as much of a concern for me as if the user scrolls up the page content is fine so ultimately it doesn't affect functionality

cssjoe (approved):

PR #882, opened 2024-05-07
Always Cached

jacobd91:


cssjoe (changes requested):
- Fix the CSS issue for the settings page buttons.
- Use a translation function (`_x()`) for the human time diff.
- I didn't purge anything and I have 4 pending and 1 postponed, but nothing listed.  It may be due to switching branches since the previous tests on this PR.
- I changed the WP Cron from weekly to daily and the event schedule didn't change.  It only updated WP Cron if I disable and then enable the setting.
- The settings page title says only "Extensions", plugin name, and the site title.  The title should change to indicate the settings page, for example: "Extensions - Always Cached - Settings | W3 Total Cache ‹ SITE_TITLE".
- I browsed to a page in Incognito, edited it, browsed again (no changes), clicked "Clear Queue", and browsed to it again; the old content was still there -- no update.  The page should clearly explain what the actions do.  I thought that it would purge/expire the cache objects, but it appears to only remove the entries from the list.  It's ok to just remove the items from the list and leave it to the cache object expiration times.

cssjoe (approved):
The `Extension_AlwaysCached_Queue::create_table()` function is run on every W3TC admin page load and issues an SQL query to create the queue table if it doesn't exist.  I suppose it's ok, since a call would be needed to check if the table exists.
When I run `wp cron event run w3tc_alwayscached_wp_cron`, I see HTML; can be fixed another time.

PR #884, opened 2024-05-11
Upgrading JSMin.php to latest version to support ES6

alexw23:
Had a few issues with [string literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in the current version of JSMin.

Have upgraded to the latest version and the issues have gone away so would be great to get this merged in.

Source [taken from here](https://github.com/mrclay/jsmin-php/blob/master/src/JSMin/JSMin.php) and configured as per the previous file (i.e change name space and add in exceptions).

Also might fix https://github.com/BoldGrid/w3-total-cache/issues/418.


cssjoe:
We'll test this out and check compatibility with https://github.com/mrclay/jsmin-php/blob/master/CHANGELOG.md#243---2022-12-13

cssjoe (approved):

cssjoe (approved):

PR #888, opened 2024-06-05
Remove JS prompt from Remove CSS/JS

jacobd91:


cssjoe (changes requested):
When I click the "Add" button, the page scrolls up too far and the input field in focus is behind the header.

When I add a new target block and change focus, I get a JS prompt.  I can't even click a link on the page until I type a character in the input field.  It would be better to add a red border around the input field and/or a warning message under the field.  If a Target CSS/JS field is empty and the settings saved, then it should just be skipped and not saved.

When I add a target block and click the Delete button, it does not delete; I get a JS prompt because the field is empty.  I should be able to delete the empty block without issue.

If I click the Delete button, the block should be removed and not display a JS prompt to confirm.

There are several instances of "ID" in variable names. https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#abbreviations-and-acronyms says to name variables this way: "singleId", "maxId".

cssjoe (approved):

PR #893, opened 2024-06-10
Fix PHP 8 warnings for Redis connect in PHP 8.1

cssjoe:
To test:

1. Switch PHP to 8.1 with error logging enabled.
2. Configure Page Cache or Fragment Cache to use Redis.
3. GO to the Page Cache or Fragment Cache settings page.
4. Click the "Test" button.
5. See the test passed and no errors logged.

Our minimum PHP version is now 7.2.5, so we can use the [Null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op) (`??`), which was introduced in PHP 7.0.


jacobd91 (approved):
Tested and approved

PR #895, opened 2024-06-11
Updated Azure dependencies

cssjoe:
I removed the old Azure library directory and installed the dependencies via Composer.  I tested Microsoft Azure using a free account.  I configured the storage account in W3TC and tested operations under PHP 8.3.

To test:
1. Create a free Azure account
2. Create a storage account
3. Configure the CDN push engine in W3TC
4. Create a storage container on the W3TC CDN settings page
5. Test the configuration using the test button
6. Upload files, such as the media library


jacobd91 (approved):
Approved

PR #896, opened 2024-06-11
Added proof block for REST API caching

jacobd91:
This PR adds a proof block for the REST API Caching feature under Page Cache -> REST API section

Additionally this PR adds the option to add score blocks to radiogroups if values are configured as pro, as well as making the score block label configurable


jacobd91:
image

cssjoe (approved):

PR #898, opened 2024-06-13
Fix Strings translation in two PHP files

DAnn2012:
Fix Strings translation in two PHP files


cssjoe (approved):

cssjoe:
@DAnn2012 Thank you for your contribution.  We'll include your changes in the next patch release.

PR #899, opened 2024-06-13
Update StackPath message and CDN default to none

cssjoe:
To test:
1. Select StackPath as the CDN and save settings.
2. See the notice above the selection.

New installations will default to no CDN selected.

Ref: https://github.com/BoldGrid/w3-total-cache-internal/issues/1814


jacobd91 (approved):
Tested and approved

PR #911, opened 2024-07-02
Fix Remove CSS/JS URL matching

cssjoe:
There was an issue matching partial/relative URLs/URIs.  I created a new static method instead of using `array_intersect()`.  I also updated the array of "current_pages".


jacobd91 (approved):
Approved

jacobd91 (dismissed):

jacobd91 (approved):
Approved

PR #916, opened 2024-07-11
Fix WP-CLI option output and more

### LINKED ISSUE #31: Adjust nested json settings from wp-cli
As it stands, you can enable most caching types from the Cli, in a format similar to:
`wp w3-total-cache option set pgcache.enabled true --type=boolean`

However, as far as I can tell, there is not a way to configure fragment cache options using this method, as fragment cache settings are stored as a JSON object.
```# wp w3-total-cache option get fragmentcache  --type=array
{
    "engine": "file"
}
```

However there is no convenient way to serialize back to the JSON object, the closest serializer is the "array" serializer, which is still incredibly inappropriate:

```# wp w3-total-cache option set fragmentcache '{"engine": "file"}'  --type=array
Success: Option updated successfully 
# wp w3-total-cache option get fragmentcache engine --type=array
[
    "{\"engine\": \"file\"}"
]
```

Ideally the solution would be to allow serialization of nested json data, to some degree. Something such as `--type=object`, or allowing name selection such as `w3-total-cache option set fragmentcache["engine"] file`

Happy to contribute myself if I am correct that this is indeed a needed feature.

cssjoe:
- Fixed the WP-CLI option output to escape properly.
- Added a new "json" type for the "option get" and "option set" that outputs valid JSON.
- Applied PHPCS WPCS and formatting.
- Added "w3tc" as a WP-CLI command alias.

Fixes #31

Example commands:
```
wp w3tc export export.json
wp w3tc export export.json --mode=0400
wp w3tc option set test '{"testing": true, "extra": "stuff..."}' --type=json
wp w3tc option get test --type=json
```

### DIFF
--- a/Cli.php
+++ b/Cli.php
@@ -376,6 +376,62 @@ class W3TotalCache_Command extends \WP_CLI_Command {
 	}
 
 	/**
+	 * Export configuration file
+	 *
+	 * ## OPTIONS
+	 * <filename>
+	 * : Filename to export -- Sanitized with sanitize_file_name()
+	 *
+	 * [--mode=<mode>]
+	 * : Mode of the file. Default: 0600 (-rw-------)
+	 *
+	 * @global $wp_filesystem
+	 * @see get_filesystem_method()
+	 *
+	 * @param array $args Arguments.
+	 * @param array $vars Variables.
+	 * @throws \Exception Exception.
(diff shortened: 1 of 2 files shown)


jacobd91 (approved):
Approved

PR #917, opened 2024-07-12
[DRAFT] Adding Managed Identity support for Microsoft Azure Storage

ZubaeyrMSFT:
- Added a new engine type for Microsoft Azure Storage with Managed Identity ('azuremi')
- Using REST based service instead of relying on Microsoft Azure Storage PHP SDK.

w3tc_mi1
w3tc_mi2


inc/options/cdn/azuremi.php
```diff
@@ -0,0 +1,122 @@
...
+?>
+<tr>
+	<th style="width: 300px;"><label for="cdn_azuremi_user"><?php esc_html_e( 'Account name:', 'w3-total-cache' ); ?></label></th>
+	<td>
+		<input id="cdn_azuremi_user" class="w3tc-ignore-change" type="text"
+			<?php Util_Ui::sealing_disabled( 'cdn.' ); ?> name="cdn__azuremi__user" value="<?php echo esc_attr( $this->_config->get_string( 'cdn.azuremi.user' ) ); ?>" size="30" />
+	</td>
+</tr>
+<tr>
+	<th><label for="cdn_azuremi_clientid"><?php esc_html_e( 'Entra client id:', 'w3-total-cache' ); ?></label></th>
```
BlackRider97, line 16: `Entra Client ID`

CdnEngine_Azure_MI.php
```diff
@@ -0,0 +1,414 @@
...
+	 * @return bool
+	 */
+	public function _init( &$error ) {
+		if ( empty( $this->_config['user'] ) ) {
+			$error = 'Empty account name.';
+			return false;
+		}
+
+		if ( empty( $this->_config['clientid'] ) ) {
+			$this->_config['clientid'] = getenv('ENTRA_CLIENT_ID');
```
BlackRider97, line 50: this mean that it can be passed as Environment variable also? 

How as user will specify these Environment variables from Plug-in experience?

CdnEngine_Azure_MI.php
```diff
@@ -0,0 +1,414 @@
...
+			$error = 'Empty container name.';
+
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Uploads files to S3
```
BlackRider97, line 67: upload to Microsoft Azure Blob Storage

CdnEngine_Azure_MI.php
```diff
@@ -0,0 +1,414 @@
...
+					sprintf( 'Unable to delete blob (%s).', $exception->getMessage() ),
+					$file );
+			}
+		}
+
+		return !$this->_is_error( $results );
+	}
+
+	/**
+	 * Tests S3
```
BlackRider97, line 196: update `S3` string

CdnEngine_Azure_MI.php
```diff
@@ -0,0 +1,414 @@
...
+		}
+
+		$string = 'test_azure_' . md5( time() );
+
+		if ( !$this->_init( $error ) ) {
+			return false;
+		}
+
+		try {
+			$containers = CdnEngine_Azure_MI_Utility::listContainers( $this->_config['clientid'],
```
BlackRider97, line 213: You can use single container GET API instead of LIST API here 
https://learn.microsoft.com/en-us/rest/api/storageservices/get-container-properties?tabs=microsoft-entra-id

CdnEngine_Azure_MI_Utility.php
```diff
@@ -0,0 +1,352 @@
...
+}
+
+/**
+ * 
+ * This class defines utility functions for Azure blob storage access using Managed Identity.
+ * @author  Zubair <zmohammed@microsoft.com>
+ */
+class CdnEngine_Azure_MI_Utility {
+
+    const ENTRA_API_VERSION = '2019-08-01';
```
BlackRider97, line 16: these API versions are latest recommended versions?

Cdn_Plugin_Admin.php
```diff
@@ -191,6 +191,11 @@ public function w3tc_settings_general_boxarea_cdn() {
 			'optgroup' => $optgroup_push,
 		);
 
+		$engine_values['azuremi'] = array(
+			'label'    => \__( 'Microsoft Azure Storage (Managed Identity)', 'w3-total-cache' ),
```
BlackRider97, line 195: Microsoft Azure Storage with Managed Identity (Recommended) ?

CdnEngine_Azure_MI_Utility.php
cssjoe: I see that as the only value in the [API documentation](https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#rest-endpoint-reference).

inc/options/cdn/azuremi.php
ZubaeyrMSFT: Changed it to `Entra client ID`. I see that sentence case is being used consistently all across.

CdnEngine_Azure_MI.php
ZubaeyrMSFT: This is an optional step that can help get the value from the env variable if the value is not available in the input config. Would be helpful with App Service if there is any issue while updating the plugin configuration.

CdnEngine_Azure_MI.php
ZubaeyrMSFT: Done.

CdnEngine_Azure_MI.php
ZubaeyrMSFT: Done.

Cdn_Plugin_Admin.php
ZubaeyrMSFT: We cannot add (Recommended) suffix here because there are so many other options like S3, Google drive etc., It would make it look like that Azure Blob Storage is recommended among all the content distribution options. The change might get rejected.

Please do let me know what you think?

CdnEngine_Azure_MI.php
ZubaeyrMSFT: We can probably take this up in next versions. This is anyways only for 'Test' button functionality and not for actual operations. 

The reason is that I am trying to keep code changes to minimum and just replacing the calls to blob storage functions while keep the rest of the code and core logic same as it was before.

cssjoe:
This work is being done on the `azure-mi` branch.

PR #919, opened 2024-07-12
Add image service limits and update transient usage

cssjoe:
You can now get the Image Service usage and limit data from `Extension_ImageService_Plugin::get_api()->get_usage();`.  If you need fresh data, then pass a boolean: `Extension_ImageService_Plugin::get_api()->get_usage( true )`.
The data is now stored in a transient for 1 day.  The settings page gets transient data if it exists.


jacobd91 (approved):
Approved

PR #944, opened 2024-09-05
Update support link color in admin bar and menu

avonville:
Fixes [internal-1816](https://github.com/BoldGrid/w3-total-cache-internal/issues/1816)

### Testing

- Suppot link in the admin menu is now blue instead of red.

### DIFF
--- a/pub/css/options.css
+++ b/pub/css/options.css
@@ -734,16 +734,12 @@ td.w3tc_config_value_text {
 	margin-right: auto;
 }
 
-.w3tc_menu_item_pro {
+.w3tc_menu_item_pro, #toplevel_page_w3tc_dashboard li a[href="admin.php?page=w3tc_support"] {
 	color: #0FA9B5;
 }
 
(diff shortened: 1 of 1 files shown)


cssjoe (changes requested):
Change the hover color to white like it is for "Fragment Cache".  It may be best to just apply the "w3tc_menu_item_pro" class, which already has a hover definition.  That way you don't need the ugly "#toplevel_page_w3tc_dashboard li a[href="admin.php?page=w3tc_support"]" selector definition.

avonville:
> Change the hover color to white like it is for "Fragment Cache". It may be best to just apply the "w3tc_menu_item_pro" class, which already has a hover definition. That way you don't need the ugly "#toplevel_page_w3tc_dashboard li a[href="admin.php?page=w3tc_support"]" selector definition.

added the CSS class to `.w3tc_menu_item_pro` CSS block... unable to resolve the comment as its not part of a review. Re-requested review from joe

PR #968, opened 2024-09-27
Add premium service tab

avonville:
Adds a premium service tab to the core cache types.

### Testing

1. Navigate to the general settings page.
2. Tabs have been added to: page cache, browser, object, db, minify, and cdn


inc/options/general.php
```diff
@@ -317,7 +347,8 @@
 			),
 			'',
 			'database_cache',
-			Util_UI::admin_url( 'admin.php?page=w3tc_dbcache' )
+			Util_UI::admin_url( 'admin.php?page=w3tc_dbcache' ),
+			'w3tc_premium_services'
```
cssjoe, line 351: The Premium Services div is missing around line 373.  Since you have duplicate content, use a new function or view file to require and reuse it.

cssjoe (changes requested):
The Premium Services div is missing for Database Cache.
Since you have duplicate content, use a new function or view file to require and reuse it.

cssjoe (changes requested):
- The Database Cache section is missing the div/content for Premium Services.
- Minify may have the incorrect content -- "can setup page cache for you"
- Object Cache may have the incorrect content -- "can setup page cache for you"
- Browser Cache may have the incorrect content -- "can setup page cache for you"
- Use "set up" instead of "setup" -- see https://grammarist.com/spelling/set-up-vs-setup/

inc/options/general.php
```diff
@@ -411,6 +420,8 @@
 			?>
 		</table>
 
+		<?php echo wp_kses_post( Util_Ui::get_premium_service_tab( 'database_cache' ) ); ?>
```
cssjoe, line 423: Incorrect tab key; use "object_cache".

cssjoe (changes requested):
- The Object Cache tab has content for Database Cache.
- Add `@since X.X.X` to all new method doc blocks.  It's updated before publishing releases.
- If possible (recommended), use argument and return type declarations -- see https://www.php.net/manual/en/language.types.declarations.php

cssjoe (approved):

PR #970, opened 2024-10-01
Add hooks for rules customization

anatolinicolae:
In some cases there's need for more surgical customizations of the rules in an .htaccess file. Often it's also best not to repeat ourselves, adding the same rules over and over instead of relying on centralized solutions ([reference](https://docs.apiscp.com/admin/Benchmarking/#trimming-htaccess)).

This PR adds more hooks for third-party plugins to hook into, and manipulate Apache rules as needed.

### DIFF
--- a/BrowserCache_Environment.php
+++ b/BrowserCache_Environment.php
@@ -249,7 +249,7 @@ class BrowserCache_Environment {
 		$rules = '';
 
 		if ( count( $mime_types ) ) {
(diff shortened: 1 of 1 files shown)


cssjoe:
@anatolinicolae I created https://github.com/BoldGrid/w3-total-cache/pull/980 to replace this suggested pull request, which implements more filters and corrects some coding standards violations.

PR #981, opened 2024-10-29
Settings not saved in the Cloudflare extension

### LINKED ISSUE #943: Settings not saved in the Cloudflare extension
In the W3TC Cloudflare extension - Performance>Extensions>Cloudflare - Settings, once the checkboxes for **Flush Cloudflare on Post Modifications** and **Exclude minified JS files from being processed by Rocket Loader** are checked, and save all settings, those settings are not saved and the checkboxes remain unchecked.

image

In the Performance>General Settings, under the Exclude minified JS files from being processed by Rocket Loader section the **Flush Cloudflare on Post Modifications** checkbox can be checked and settings saved.

image

avonville:
Fixes #943


cssjoe (changes requested):
- The link to General Settings needs to be updated -- see comment.
- Not required, but you may want to add an "Advanced Settings" tab on the General Settings Cloudflare section to go to the extension settings page.

cssjoe (approved):

PR #1008, opened 2024-12-06
Settings Tab GA Updates

avonville:
Removed unnecessary jQuery and updated the eventCategory value to have the section of the settings the tab where it was clicked i.e. Page Cache, Minify, Object Cache, etc.

### DIFF
--- a/Generic_Plugin_Admin.php
+++ b/Generic_Plugin_Admin.php
@@ -492,10 +492,10 @@ class Generic_Plugin_Admin {
 
 
 				// Track clicks on W3TC Pro Services tab.
-				document.addEventListener('click', function( event ) {
-					if ( event.target.getAttribute( 'data-tab-type') ) {
+				document.addEventListener('click', function(event) {
+					if (event.target.getAttribute('data-tab-type')) {
 						w3tc_ga('event', 'click', {
-							'eventCategory': event.target.closest(".postbox-tabs").getAttribute("id"),
+							'eventCategory': event.target.closest('.postbox-tabs').getAttribute('id'),
 							'eventLabel': event.target.getAttribute('data-tab-type'),
 						});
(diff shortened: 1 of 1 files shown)


Generic_Plugin_Admin.php
```diff
@@ -492,10 +492,10 @@ function w3tc_ga(){dataLayer.push(arguments);}
 
 
 				// Track clicks on W3TC Pro Services tab.
-				document.addEventListener('click', function(event) {
-					if ( jQuery( event.target ).attr( 'data-tab-type') ) {
+				document.addEventListener('click', function( event ) {
+					if ( event.target.getAttribute( 'data-tab-type') ) {
 						w3tc_ga('event', 'click', {
-							'eventCategory': 'w3tc-tabs',
+							'eventCategory': event.target.closest(".postbox-tabs").getAttribute("id"),
```
bwmarkle, line 498: The code is looking good overall, and I appreciate the clear logic you've implemented for tracking clicks on the Pro Services tab.  

I noticed a couple of style inconsistencies that we should address to ensure the code aligns with our standards:  

### 1. **Inconsistent Use of Quotes**  
There’s a mix of single (`'`) and double (`"`) quotes being used. For example:  

- `'data-tab-type'` (single quotes)  
- `"id"` (double quotes)  

To maintain consistency, it’s best to stick to one style throughout the file. **Single quotes** are generally preferred in JavaScript unless there’s a specific reason to use double quotes (e.g., if the string contains single quotes).  

### 2. **Spaces Around Quotes**  
There’s a spacing inconsistency in cases like this:  

```javascript
( 'data-tab-type')

bwmarkle (changes requested):

cssjoe (approved):

PR #1016, opened 2024-12-16
Fragment Cache Frontend Fix

jacobd91:


jacobd91:
Todo
- Fix general setting page block showing advanced tab when not pro but active
- Fix on pro loss the extension remains active (would fix the above issue)
- Add disk usage warning

cssjoe (approved):

PR #1019, opened 2024-12-19
updated api to admin-ajax

avonville:
Updated the Forums API used on the General Setting page to be a PHP function to get data.


cssjoe (changes requested):
I was getting timeouts for some calls, so I increased the timeout to 10 seconds.  The JS needs to handle such an error.  When a timeout occurred, the spinner remained.  An error message should be printed and maybe a retry button.
Also, use translated text passed to JS via script localization.

PR #1027, opened 2025-01-07
fix(Cdn_Plugin.php): improve srcset regex pattern to handle multiline attributes

timohubois:
Add PCRE_DOTALL flag to srcset regex pattern to properly match multiline srcset attributes in HTML content. This ensures CDN replacement works correctly even when srcset values span multiple lines.


cssjoe (approved):

cssjoe:
@timohubois Thank you for your contribution.

PR #1031, opened 2025-01-13
Bunny CDN Widget Recommend Notice

jacobd91:
This PR simply updates the unauthorized Bunny CDN widget to show an admin notice indicating that another CDN was detected but that we recommend Bunny CDN instead

To test simply configure a non Bunny CDN and visit the dashboard. The Bunny CDN widget should show a notice recommending Bunny CDN

### DIFF
--- a/Cdn_BunnyCdn_Widget_View_Unauthorized.php
+++ b/Cdn_BunnyCdn_Widget_View_Unauthorized.php
@@ -103,7 +103,7 @@ defined( 'W3TC' ) || die();
 				'cdn.bunnycdn.widget.v2.no_cdn',
 				\sprintf(
 					// translators: 1 configured CDN name, 2 HTML acronym for Content Delivery Network (CDN).
-					\__( 'W3 Total Cache has detected that you have the %1$s %2$s configured which is fine, but we recommend using BunnyCDN instead.', 'w3-total-cache' ),
+					\__( 'W3 Total Cache has detected that you are using the %1$s %2$s, which is fully supported and compatible. For optimal performance and value, we recommend considering BunnyCDN as an alternative.', 'w3-total-cache' ),
 					$cdn_name,
 					'<acronym title="' . \__( 'Content Delivery Network', 'w3-total-cache' ) . '">' . \__( 'CDN', 'w3-total-cache' ) . '</acronym>'
 				)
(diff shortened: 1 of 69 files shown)


Cdn_BunnyCdn_Widget_View_Unauthorized.php
```diff
@@ -13,7 +14,105 @@
 ?>
 
 <div id="bunnycdn-widget" class="w3tc_bunnycdn_signup">
-	<?php if ( ! $c->get_boolean( 'cdn.enabled' ) ) : ?>
+	<?php
+	$cdn_engine = $c->get_string( 'cdn.engine' );
+	$cdn_name   = '';
```
cssjoe, line 19: Get the CDN name from https://github.com/BoldGrid/w3-total-cache/blob/2.8.2/Cache.php#L103 and not duplicate code using a new switch/case statement.

Cdn_BunnyCdn_Widget_View_Unauthorized.php
```diff
@@ -13,7 +14,105 @@
 ?>
 
 <div id="bunnycdn-widget" class="w3tc_bunnycdn_signup">
-	<?php if ( ! $c->get_boolean( 'cdn.enabled' ) ) : ?>
+	<?php
+	$cdn_engine = $c->get_string( 'cdn.engine' );
```
cssjoe, line 18: The site may have Bunny CDN configured as full-site delivery, so check both settings (enabled and engine) to see and only display the notice if not already using Bunny for one of them.

cssjoe (changes requested):
See comments.

Cdn_BunnyCdn_Widget_View_Unauthorized.php
jacobd91: Changed to use Cache::engine_name and updated logic to check if cdn_engine is not empty so that the notice doesn't display as "detected none"

Cdn_BunnyCdn_Widget_View_Unauthorized.php
jacobd91: Updated to take into account CDN enabled and engine selection in case of bunny

cssjoe (approved):

PR #1046, opened 2025-01-28
License Verify/Apply Update

jacobd91:
This PR simply removes the `verify license key` button on the General Settings page and adds an admin notice that displays on save indicating the result of license activation

### DIFF
--- a/Licensing_Plugin_Admin.php
+++ b/Licensing_Plugin_Admin.php
@@ -137,6 +137,13 @@ class Licensing_Plugin_Admin {
 						);
 						break;
 
+					case ( strpos( $status, 'inactive.not_present' ) === 0 ):
+						$messages[] = array(
+							'message' => __( 'Your previous W3 Total Cache Pro license key was not found and cannot be deactivated.', 'w3-total-cache' ),
+							'type'    => 'info',
+						);
+						break;
+
 					case ( strpos( $status, 'inactive' ) === 0 ):
 						$messages[] = array(
 							'message' => __( 'Your previous W3 Total Cache Pro license key has been deactivated.', 'w3-total-cache' ),
@@ -149,6 +156,7 @@ class Licensing_Plugin_Admin {
 							'message' => __( 'Your previous W3 Total Cache Pro license key is invalid and cannot be deactivated.', 'w3-total-cache' ),
 							'type'    => 'error',
 						);
+						break;
 				}
 			}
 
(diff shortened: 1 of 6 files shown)


Licensing_Plugin_Admin.php
```diff
@@ -86,24 +95,33 @@ public function admin_head_licensing_upgrade() {
...
 			}
 			$changed = true;
-		} elseif ( $old_config->get_string( 'plugin.license_key' ) !== $config->get_string( 'plugin.license_key' ) ) {
-			$result = Licensing_Core::activate_license( $config->get_string( 'plugin.license_key' ), W3TC_VERSION );
-			if ( $result ) {
-				$this->site_activated = true;
+		} elseif ( '' !== $new_key && '' !== $old_key && $new_key !== $old_key ) {
+			// Current key is set and new different key provided. Deactivating old and activating new.
+			$deactivate_result = Licensing_Core::deactivate_license( $old_key );
+			$activate_result   = Licensing_Core::activate_license( $new_key, W3TC_VERSION );
```
cssjoe, line 122: `$activate_result` should be `$result` to avoid a PHP warning and preventing display of a message.
```
PHP Warning:  Undefined variable $result in Licensing_Plugin_Admin.php on line 136
```

Licensing_Plugin_Admin.php
```diff
@@ -86,24 +95,33 @@ public function admin_head_licensing_upgrade() {
...
+			if ( $deactivate_result && $activate_result ) {
```
cssjoe, line 123: `Licensing_Core::deactivate_license()` returns a boolean, but `Licensing_Core::activate_license()` does not -- it returns (on success):
```
(object) array(
   'license' => 'site_active',
   'license_status' => 'active.by_rooturi',
)
```

When a key is swapped, it doesn't show the correct messages.  Each action should show a message.  For example, messages like "Your old Pro license key is not valid, so no action was taken." and "YYour Pro license key is correct and has been applied.".

cssjoe (changes requested):
See comments.  Also,:
Use `empty( $var )` for `'' === $var` and `! empty( $var )` for `'' !== $var` in all of the checks for empty license key strings.
There are too many `if` `elseif` `else` statements.  You should convert them to `switch( true )` `case` conditions.

Licensing_Plugin_Admin.php
jacobd91: This was due to a missed reference to the old $result being missed. It should now display a message

cssjoe (approved):

PR #1047, opened 2025-01-29
Nginx Page Cache - Disk Enhanced Fix

jacobd91:
This PR addresses an issue with Page Cache when using the Disk Enhanced engine under nginx where page/posts would result in an nginx 404 instead of hitting PHP when no cache entry is found


cssjoe:
The fix was reverted.  It appears to be an Nginx environment configuration issue.  If we get any other reports, then we can revisit the issue.

PR #1063, opened 2025-02-21
Remove W3TC page_view GA event from Plugins page

jacobd91:


jacobd91:
Besides the restoring of `w3tc_ga('js', new Date());` this PR can be discarded

PR #1128, opened 2025-07-21
Dunning Notice - Add Refresh License Button

jacobd91:
This PR simply adds a button to the "past due billing" notice to allow users to force a refresh of the license state. Once refreshed the notice should no longer display

To test, first obtain a subscription for both W3TC and TCDN and then set both to expire. Once both reach a dunning state you should see 2 admin notices in W3TC admin indicating that billing is past due. These notices should contain a link to update your billing on file in Recurly as well as a button to "Refresh License Status" afterwards

Once you update your billing it should immediately (or shortly afterwards) charge and update the license/cdn status back to "active", however the plugin won't likely update it's state due to the interval being 5 days. This PR does adjust that so if the state is in dunning, the next time it checks the interval will change to 1 hour until resolved. This is where the "Refresh License Status" button comes into play as clicking it should immediately force a state refresh which should cause the dunning notices to no longer display


Licensing_Plugin_Admin.php
```diff
@@ -354,14 +354,16 @@ private function get_cdn_notice( $status, $license_key ) {
 				$billing_url = $this->get_billing_url( $license_key );
 				if ( $billing_url ) {
 					return sprintf(
-						// Translators: 1 Total CDN tradmark, 2 opening HTML a tag to billing portal, 3 closing HTML a tag.
+						// Translators: 1 Total CDN tradmark, 2 opening HTML a tag to billing portal
+						// Translators: 3 closing HTML a tag, 4 button to refresh license status.
 						__(
-							'Your %1$s subscription payment is past due. Please update your %2$sBilling Information%3$s to prevent service interruption',
+							'Your %1$s subscription payment is past due. Please update your %2$sBilling Information%3$s to prevent service interruption. Once your billing information has been updated and payment is successful please manually %4$s otherwise it may take up to 1 hour to refresh on it\'s own.',
```
jamesros161, line 360: Grammer: This should be its own

jamesros161: Grammer: This should be its own not it's own

pub/js/options.js
```diff
@@ -1690,6 +1690,35 @@ jQuery(function() {
...
+		jQuery.ajax({
+			url: ajaxurl,
+			type: 'POST',
+			data: {
+				action: 'w3tc_ajax',
+				_wpnonce: nonce,
+				w3tc_action: action
+			},
+			success: function(response) {
+				console.log('License refresh complete', response);
```
jamesros161: I understand the console.error() down below, but do we need a console.log() here?

Licensing_Plugin_Admin.php
```diff
@@ -585,7 +587,24 @@ public function w3tc_notes( $notes ) {
 	 */
 	private function maybe_update_license_status( $force = false ) {
 		$state = Dispatcher::config_state();
-		if ( time() < $state->get_integer( 'license.next_check' ) && ! $force ) {
+
+		$next_check = $state->get_integer( 'license.next_check' );
+
+		/**
+		 * If the license status is in dunning and the next_check interval is greatter
```
jamesros161: Typo: greatter should be greater

Licensing_Plugin_Admin.php
jacobd91: Updated

Licensing_Plugin_Admin.php
jacobd91: Updated

pub/js/options.js
jacobd91: Probably not. Removing

PR #1135, opened 2025-07-29
Total CDN Feature Showcase

jacobd91:
To test set up your install with no CDN product or configuration

Then navigate to the Feature Showcase and you should see a new card for Total CDN that shows a link to the documentation "More Info" and a button to subscribe to Total CDN

Then purchase a CDN subscription and revisit the page

If the autoconfigure was successful you should see a settings button instead that will take you to the CDN page

If you then disable CDN on the general settings page, the settings button will instead link to the CDN section on the general settings


jamesros161:
Closed due to strange stuff from the base branch, recreated in https://github.com/BoldGrid/w3-total-cache/pull/1136

PR #1139, opened 2025-07-30
ENG5-166 manual config improvements

jamesros161:


Cdn_TotalCdn_Popup.php
```diff
@@ -288,11 +290,48 @@ public function w3tc_ajax_cdn_totalcdn_list_pull_zones() {
 		$server_ip = ! empty( $_SERVER['SERVER_ADDR'] ) && \filter_var( \wp_unslash( $_SERVER['SERVER_ADDR'] ), FILTER_VALIDATE_IP ) ?
 			\filter_var( \wp_unslash( $_SERVER['SERVER_ADDR'] ), FILTER_SANITIZE_URL ) : null;
 
+		// Determine if a new pull zone can be added.
+		$max_pull_zones    = 0;
+		$can_add_pull_zone = true;
+		if ( ! empty( $account_info['Products'] ) && \is_array( $account_info['Products'] ) ) {
```
jacobd91, line 296: Unchecked array keys could result in warnings/errors if the values aren't set? Add isset check or null coalesce ?? for safety

Cdn_TotalCdn_Api.php
```diff
@@ -219,6 +219,21 @@ public function list_pull_zones() {
...
+	 * @since x.x.x
+	 *
+	 * @link https://docs.bunny.net/reference/account_get
+	 *
+	 * @return array|WP_Error API response or error object.
+	 */
+	public function get_account() {
+		$this->api_type = 'account';
+
+		return $this->wp_remote_get( \esc_url( $this->api_base_url . '/account' ) );
```
jacobd91, line 234: Directly returning the result of wp_rmote_get without error checking or validation. The call to this doesn't appear to check either which could potentially cause warnings/errors

Cdn_TotalCdn_Popup.php
```diff
@@ -288,11 +290,48 @@ public function w3tc_ajax_cdn_totalcdn_list_pull_zones() {
...
+
+				$origin_host = \strtolower( \wp_parse_url( $pz['OriginUrl'], PHP_URL_HOST ) );
+				if ( $origin_host === $home_host ) {
+					$pull_zone_id = $pz['Id'];
+					break;
+				}
+			}
+		}
+
+		$origin_url_host = \wp_parse_url( \home_url(), PHP_URL_HOST );
```
jacobd91: This appears identical to the value on line 311. Maybe assign to a variable to reduce repeated code and improve readability?

Cdn_TotalCdn_Popup_View_Pull_Zones.php
```diff
@@ -36,7 +37,9 @@
 		<table class="form-table">
 			<tr>
 				<select id="w3tc-pull-zone-id">
-					<option value=""<?php echo empty( $details['pull_zone_id'] ) ? ' selected' : ''; ?>>Add a new pull zone</option>
+					<?php if ( false !== $details['can_add_pull_zone'] ) : ?>
```
jacobd91: is "can_add_pull_zone" always a boolean? If so you don't need the compare check and can just have

if ( $details['can_add_pull_zone'] ) :

Cdn_TotalCdn_Popup.php
```diff
@@ -288,11 +290,48 @@ public function w3tc_ajax_cdn_totalcdn_list_pull_zones() {
...
+			}
+
+			if ( 0 !== $max_pull_zones ) {
+				$can_add_pull_zone = \count( $pull_zones ) < $max_pull_zones;
+			}
+		}
+
+		// Determine default pull zone id.
+		$pull_zone_id = $config->get_integer( 'cdn.totalcdn.pull_zone_id' );
+		if ( empty( $pull_zone_id ) ) {
```
jacobd91, line 318: This block is confusing, maybe add comments explain what it does and any other details to make it clear

Cdn_TotalCdn_Popup.php
jamesros161: `! empty( $account_info['Products'] )` will return false if the array key isn't set, so that will prevent the warnings.

Cdn_TotalCdn_Api.php
jamesros161: This doesn't use wp_remote_get directly, it uses `$this->wp_remote_get()` which as a method in this class, which also uses `$this->decode_response()` which checks for error messages.

Cdn_TotalCdn_Popup_View_Pull_Zones.php
jamesros161: Changing to use `if ( ! empty( $details['can_add_pull_zone'] ) )` to cover the possibility that the array key doesn't exist

jacobd91 (approved):

PR #1167, opened 2025-08-20
ENG5-210 auto generate on save

jamesros161:
Story Ticket: [ENG5-210](https://imh-internal.atlassian.net/browse/ENG5-210)
Code Review Ticket: [ENG5-345](https://imh-internal.atlassian.net/browse/ENG5-345)
Acceptance Testing: [ENG5-346](https://imh-internal.atlassian.net/browse/ENG5-346)

[ENG5-210]: https://imh-internal.atlassian.net/browse/ENG5-210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG5-345]: https://imh-internal.atlassian.net/browse/ENG5-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG5-346]: https://imh-internal.atlassian.net/browse/ENG5-346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Extension_AiCrawler_Markdown.php
+++ b/Extension_AiCrawler_Markdown.php
@@ -295,11 +295,10 @@ class Extension_AiCrawler_Markdown {
 			return;
 		}
 
-		// Ensure that the post is not excluded.
-		// Commented out for now, as the exclusion logic is not yet implemented.
-		// if ( Extension_AiCrawler_Util::is_excluded( $post_id ) ) {
-		// 	return;
-		// }
+		// Check if the post is excluded.
+		if ( Extension_AiCrawler_Util::is_excluded( $post_id ) ) {
+			return;
+		}
 
 		// Generate markdown for the post.
 		self::generate_markdown( get_permalink( $post_id ) );
--- a/Extension_AiCrawler_Markdown.php
+++ b/Extension_AiCrawler_Markdown.php
@@ -178,6 +178,7 @@ class Extension_AiCrawler_Markdown {
 			$response = Extension_AiCrawler_Central_Api::call( 'convert', 'POST', array( 'url' => $url ) );
 
 			if ( empty( $response['success'] ) || empty( $response['data']['markdown_content'] ) ) {
+				error_log( 'Error Response: ' . json_encode( $response ) );
 				update_post_meta( $post_id, self::META_STATUS, 'error' );
 				continue;
 			}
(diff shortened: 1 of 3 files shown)


Extension_AiCrawler_Markdown.php
```diff
@@ -147,7 +163,8 @@ public static function generate_markdown( $url ) {
 		}
 
 		update_post_meta( $post_id, self::META_STATUS, 'queued' );
-		update_post_meta( $post_id, self::META_SOURCE_URL, $url );
+		update_post_meta( $post_id, self::META_SOURCE_URL, esc_url_raw( $url ) );
```
jacobd91: You don't need to esc_url_raw here as it was already done on line 158. Just pass $url

Line 158: $url     = esc_url_raw( $url );

Extension_AiCrawler_Markdown.php
```diff
@@ -195,6 +220,10 @@ public static function process_queue() {
 			update_post_meta( $post_id, self::META_MARKDOWN, $markdown );
 			update_post_meta( $post_id, self::META_MARKDOWN_URL, $markdown_url );
 			update_post_meta( $post_id, self::META_STATUS, 'complete' );
+			delete_post_meta( $post_id, self::META_ERROR_MESSAGE );
+
+			// Update the timestamp again when complete:
+			update_post_meta( $post_id, self::META_TIMESTAMP, time() );
 		}
 
 		if ( ! self::queue_has_items() ) {
```
jacobd91, line 224: At the end of process_queue you do a check for if there are queue items and if not you unschedule the cron. Is that really needed as at the start of process_queue you do the same check essentially and unschedule and exit if none are found. On plugin load you also do a check to schedule if queue items exist. Seems to me this could be optimized more?

jacobd91: Yeah, upon further review I think that first check at the start of the process_queue can be removed as the process_queue only gets scheduled if there are queue items. If at the end of processing those items and no new ones exist, it unschedules

Extension_AiCrawler_Markdown.php
```diff
@@ -238,4 +267,40 @@ public static function get_queue_posts() {
...
+	 *
+	 * @param int     $post_id Post ID.
+	 * @param WP_Post $post    Post object.
+	 * @param bool    $update  Whether this is an existing post being updated or not
+	 *
+	 * @return void
+	 */
+	public static function generate_markdown_on_save( $post_id, $post, $update ) {
+		// Avoid recursion.
+		remove_action( 'save_post', array( __CLASS__, 'generate_markdown_on_save' ), 10 );
```
jacobd91, line 279: In this context, the remove_action( 'save_post', … ) isn’t actually needed. Here’s why:

Why recursion usually happens

Recursion issues on save_post happen when your callback itself calls wp_insert_post() or wp_update_post(), which fires save_post again, re-triggering your callback in a loop.

But in your code, generate_markdown_on_save() only calls self::generate_markdown( get_permalink( $post_id ) ). That function is making an external API call and updating post meta, not the post object itself.

update_post_meta() and delete_post_meta() do not trigger save_post. They trigger updated_postmeta, but that’s separate.

Why remove_action() is unnecessary here

Since you’re not altering the post content or calling a save/update on the post itself, recursion can’t occur.

Removing the action before your logic also means you don’t re-add it in this function, so you’re relying on the action being re-added at plugin load time. That could be fragile if you ever call this function manually.

Extension_AiCrawler_Markdown.php
```diff
@@ -238,4 +267,40 @@ public static function get_queue_posts() {
...
+
+		// Don't run on autosave.
+		if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
```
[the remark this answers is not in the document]
jamesros161: This is incorrect. The action is re-added automatically during the wordpress cycle, because the save_post action is added directly in the wp_loaded hook. Each time you save a post, the wp_loaded hook is fired, causing the save_post action to be readded.

Extension_AiCrawler_Markdown.php
jamesros161: We discussed this yesterday, and decided that we should leave the remove_action() call here, just in case we make future changes to the process, and prevent issues.

jacobd91 (approved):

PR #1188, opened 2025-09-02
Fix auto-config notice

cssjoe:
Show notice only on W3TC pages.  Add return type declarations.  Fix doc blocks.

### DIFF
--- a/Cdn_TotalCdn_Auto_Configure.php
+++ b/Cdn_TotalCdn_Auto_Configure.php
@@ -496,7 +496,7 @@ class Cdn_TotalCdn_Auto_Configure {
 		// Display notice only if on a W3TC page.
 		$page = \filter_input( INPUT_GET, 'page', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
 
-		if ( empty( $page ) || ! \preg_match( '/^w3tc_/', $page ) ) {
+		if ( empty( $page ) || strpos( $page, 'w3tc_' ) !== 0 ) {
 			return false;
 		}
 


Cdn_TotalCdn_Auto_Configure.php
```diff
@@ -476,14 +486,20 @@ public static function admin_notices() {
...
+	public static function maybe_show_auto_config_notice( $cdn_enabled, $cdn_engine, $api_key, $cdn_status ): bool {
+		// Display notice only if on a W3TC page.
+		$page = \filter_input( INPUT_GET, 'page', FILTER_SANITIZE_FULL_SPECIAL_CHARS );
+
+		if ( empty( $page ) || strpos( $page, 'w3tc_' ) !== 0 ) {
+			return false;
+		}
+
 		// If the CDN is enabled and the engine is set, do not show the notice.
 		if ( $cdn_enabled && 'totalcdn' === $cdn_engine ) {
```
jacobd91, line 504: There's a bug with this I've noticed where if the checkout is successful and the auto-configure successfully applies the settings but fails to generate a pull zone it shows the notice despite the auto-configure failing. During initial testing my licensing was pointed at staging CDNAPI but my website was pointed at my dev CDNAPI so the request to generate a pull zone failed

Maybe add a check to see if the pullzone exists and is authenticated, or something?

jacobd91 (changes requested):

Cdn_TotalCdn_Auto_Configure.php
cssjoe: I recommend a separate task for additional checks.  This PR is mainly to prevent the notice from displaying on non-W3TC pages.

PR #1209, opened 2025-10-02
Sanitize mfunc comments

cssjoe:
https://imh-internal.atlassian.net/browse/ENG5-524

To test:
1. Enable mfunc -- see [how-does-page-fragment-cache-work](https://github.com/BoldGrid/w3-total-cache/wiki/FAQ:-Developers#how-does-page-fragment-cache-work)
1. Ensure that Page Cache is enabled.
1. Create a test page with HTML to utilize the feature.
1. In Incognito, browse to the page and review the source.  The literal comment/code added should not be visible.
1. View the page via REST API - "/wp-json/wp/v2/pages/{post_id}".  The comment/code should be stripped out.
1. Change the branch to the latest tag and reload the REST API page.  You will see the comment/code.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,6 +21,10 @@ This project is a WordPress plugin designed to enhance website performance throu
 - WordPress Documentation Standards for JavaScript: https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/
 
 ## Contribution Process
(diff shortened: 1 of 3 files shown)


pixel-paul:
Any chance of showing why this change was made please?

It's really not been thought through and is a breaking change that should not have been merged into a point release.

PR #1212, opened 2025-10-06
Added FSD SSL test

jacobd91:
This PR adds the SSL status test for the CDN FSD status section

To test setup with a pro license with total CDN and enable/configure FSD. This should set the hostname on save. Then you will need your DNS pointed at the CDN

Then go to the CDN settings page and click on the Check Status button within the FSD box. It should pass. To see the failure state modify Cdnfsd_TotalCdn_Status_Ssl.php to manually trigger a failure

If you remove the ssl from the pull zone manually and then perform the check, it should re-add it and pass the test

### DIFF
--- a/Cdn_TotalCdn_Api.php
+++ b/Cdn_TotalCdn_Api.php
@@ -186,6 +186,7 @@ class Cdn_TotalCdn_Api {
 	 * @link https://developer.wordpress.org/reference/hooks/http_request_timeout/
 	 *
 	 * @param float $timeout_value Time in seconds until a request times out. Default 5.
+	 *
 	 * @return float The adjusted timeout time.
 	 */
 	public function filter_timeout_time( $timeout_value ): float {
@@ -258,14 +259,15 @@ class Cdn_TotalCdn_Api {
 	 *
 	 * @since x.x.x
 	 *
-	 * @param int $id The pull zone ID.
-	 *
 	 * @link https://docs.bunny.net/reference/pullzonepublic_index2
+	 *
+	 * @param int $id The pull zone ID.
 	 *
 	 * @return array|WP_Error API response or error object.
 	 */
 	public function get_pull_zone( $id ) {
 		$this->api_type = 'account';
+		$id             = empty( $id ) ? $this->pull_zone_id : $id;
 
 		return $this->wp_remote_get(
 			\esc_url( $this->api_base_url . '/pullzone/' . $id )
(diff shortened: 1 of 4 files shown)


Cdn_TotalCdn_Api.php
```diff
@@ -272,6 +272,25 @@ public function get_pull_zone( $id ) {
...
+	 *
+	 * @param int $id The pull zone ID.
+	 *
+	 * @link https://cdn-api-dev-joec.boldgrid.com/docs/api#/operations/pullzone.get_from_provider
+	 *
+	 * @return array|WP_Error API response or error object.
+	 */
+	public function get_pull_zone_from_provider( $id ) {
+		$this->api_type = 'account';
+
```
cssjoe, line 291: Add the reverse logic for the pull zone ID from https://github.com/BoldGrid/w3-total-cache/blob/cb6273e128a4436e7ec52f0fac0ff6566b883cf9/Cdn_TotalCdn_Api.php#L350
```
$id             = empty( $id ) ? $this->pull_zone_id : $id;
```

Cdn_TotalCdn_Api.php
```diff
@@ -272,6 +272,25 @@ public function get_pull_zone( $id ) {
 		);
 	}
 
+	/**
+	 * Gets the details of a specific pull zone from the provider.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param int $id The pull zone ID.
```
cssjoe: Move `@param` line to be directly above the `@return` line.

Cdnfsd_TotalCdn_Page.php
```diff
@@ -25,8 +25,9 @@ class Cdnfsd_TotalCdn_Page {
 	public static function w3tc_ajax() {
 		$instance = new self();
 
-		\add_filter( 'w3tc_fsd_totalcdn_dns', array( '\W3TC\Cdnfsd_TotalCdn_Status_Dns', 'test_dns_status' ), 10 );
 		\add_filter( 'w3tc_fsd_totalcdn_hostname', array( '\W3TC\Cdnfsd_TotalCdn_Status_Hostname', 'test_hostname_status' ), 10 );
+		\add_filter( 'w3tc_fsd_totalcdn_dns', array( '\W3TC\Cdnfsd_TotalCdn_Status_Dns', 'test_dns_status' ), 10 );
```
cssjoe: `add_filter`'s 3rd argument defaults to `10`, so no need to specify it.

Cdnfsd_TotalCdn_Status_Ssl.php
```diff
@@ -0,0 +1,101 @@
...
+			);
+		}
+
+		$has_ssl = false;
+
+		$hostnames = isset( $response['Hostnames'] ) && is_array( $response['Hostnames'] ) ? $response['Hostnames'] : array();
+
+		foreach ( $hostnames as $entry ) {
+			$value = isset( $entry['Value'] ) ? strtolower( (string) $entry['Value'] ) : '';
+			$cert  = isset( $entry['HasCertificate'] ) ? (bool) $entry['HasCertificate'] : false;
```
cssjoe: Don't cast to "bool".  The value should already be a boolean.  In the condition below, compare as `true === $cert`.

Cdn_TotalCdn_Api.php
```diff
@@ -258,29 +259,50 @@ public function get_user() {
...
 	 * @since x.x.x
 	 *
-	 * @param array $data Data for the new pull zone.
+	 * @link https://cdn-api-dev-joec.boldgrid.com/docs/api#/operations/pullzone.get_from_provider
+	 *
+	 * @param int $id The pull zone ID.
+	 *
+	 * @return array|WP_Error API response or error object.
+	 */
+	public function get_pull_zone_from_provider( $id ) {
```
cssjoe: Make the `$id` argument optional -- `?int $id = null`.
Also, update the comment to show optional and type "?int".

cssjoe (approved):

PR #1225, opened 2025-10-17
Add support link to DNS status check failure message

jacobd91:
This PR simply adds a link to the error message if the DNS status check comes back as not pointing to the CDN

The link won't work as the article is still a draft, but the slugs have been synced so once it's published the link should work correctly. Here is the link to the draft article https://www.boldgrid.com/support/?p=164753&preview=true

### DIFF
--- a/Cdnfsd_TotalCdn_Page.php
+++ b/Cdnfsd_TotalCdn_Page.php
@@ -241,7 +241,7 @@ class Cdnfsd_TotalCdn_Page {
 			return \wp_kses(
 				\sprintf(
 					/* translators: 1: Total CDN test title. 2: Failure reason. */
-					__( '%1$s: %2$s', 'w3-total-cache' ),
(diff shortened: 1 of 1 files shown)


Cdnfsd_TotalCdn_Status_Dns.php
```diff
@@ -68,8 +68,13 @@ public static function test_dns_status() {
...
 				'message' => \sprintf(
 					// Translators: 1 host name.
-					\__( '%1$s is not pointed to a CDN provider', 'w3-total-cache' ),
-					$hostname
+					\__(
+						'%1$s is not pointed to a CDN provider.  For more information on how to configure your DNS please click %2$shere%3$s.',
+						'w3-total-cache'
+					),
+					$hostname,
+					'<a href="https://www.boldgrid.com/support/w3-total-cache/total-cdn-dns-setup/" target="_blank">',
```
cssjoe, line 76: The link is being stripped when the notice is printed.

PR #1234, opened 2025-10-28
Bunny/Total configuration conflict check

jacobd91:
This PR does 2 things

1. It fixes the existing conflict check for Bunny so you can't configure both CDN and FSD with bunny at the same time. The check had a bug where if you have both disabled and try to enable both, it would block the second enable and show a warning. If you enable 1 and save, on page load if you then try to enable the 2nd it allowed it. This PR fixes that bug
2. It adds the same check for Total CDN as to prevent users from configuring CDN and FSD with Total CDN

One thing to note is that the conflict check prevents mixing Bunny and Total for CDN/FSD

To test simply visit the General Settings page and attempt to configure. On conflict it should revert the last changed setting and show a warning

- Bunny CDN | Bunny FSD
- Total CDN | Total FSD
- Bunny CDN | Total FSD
- Total CDN | Bunny FSD

Lastly the conflict check should prevent the Total FSD confirmation popup from executing if the configuration is in conflict

### DIFF
--- a/Generic_Plugin_Admin.php
+++ b/Generic_Plugin_Admin.php
@@ -196,13 +196,36 @@ class Generic_Plugin_Admin {
 			wp_clear_scheduled_hook( 'w3tc_purge_all_wpcron' );
 		}
 
-		if ( Cdn_TotalCdn_CustomHostname::should_remove_on_save( $new_config, $old_config ) ) {
+		$skip_totalcdn_fsd_operations = false;
+
+		if ( $this->should_auto_configure_totalcdn_fsd( $new_config ) ) {
+			$auto_configurator = new Cdn_TotalCdn_Auto_Configure( $new_config );
+			$auto_result       = $auto_configurator->run( false );
+
+			if ( false === $auto_result['success'] ) {
+				$data['response_errors'][]    = 'cdn_totalcdn_fsd_auto_config_failed';
+				$skip_totalcdn_fsd_operations = true;
+
+				if ( $old_config ) {
+					$new_config->set( 'cdnfsd.enabled', $old_config->get_boolean( 'cdnfsd.enabled' ) );
+					$new_config->set( 'cdnfsd.engine', $old_config->get_string( 'cdnfsd.engine' ) );
+				} else {
+					$new_config->set( 'cdnfsd.enabled', false );
+
+					if ( 'totalcdn' === $new_config->get_string( 'cdnfsd.engine' ) ) {
+						$new_config->set( 'cdnfsd.engine', '' );
+					}
+				}
+			}
+		}
+
+		if ( ! $skip_totalcdn_fsd_operations && Cdn_TotalCdn_CustomHostname::should_remove_on_save( $new_config, $old_config ) ) {
 			$result = Cdn_TotalCdn_CustomHostname::remove( $new_config, $old_config );
 
 			if ( empty( $result['success'] ) && empty( $result['skipped'] ) ) {
 				$data['response_errors'][] = 'cdn_totalcdn_fsd_custom_hostname_remove_failed';
 			}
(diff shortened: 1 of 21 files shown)


Generic_Plugin_Admin.php
```diff
@@ -1151,6 +1159,9 @@ public function admin_notices() {
 				'<strong style="color:#f00;">require_once(ABSPATH . \'wp-settings.php\');</strong>'
 			),
 			'pull_zone'                               => __( 'Pull Zone could not be automatically created.', 'w3-total-cache' ),
+			'cdn_fsd_conflict_bunnycdn'               => esc_html__( 'Bunny CDN cannot be enabled for both CDN and Full Site Delivery. Please disable one and save again.', 'w3-total-cache' ),
+			'cdn_fsd_conflict_totalcdn'               => esc_html__( 'Total CDN cannot be enabled for both CDN and Full Site Delivery. Please disable one and save again.', 'w3-total-cache' ),
```
cssjoe: Use `W3TC_CDN_NAME` instead of "Total CDN".

Generic_Plugin_Admin.php
```diff
@@ -1151,6 +1159,9 @@ public function admin_notices() {
...
+			'cdn_fsd_conflict_mixed'                  => esc_html__( 'Bunny CDN and Total CDN cannot be enabled across CDN and Full Site Delivery at the same time. Please disable one and save again.', 'w3-total-cache' ),
```
cssjoe: Use `W3TC_CDN_NAME` instead of "Total CDN".

Generic_Plugin_Admin.php
```diff
@@ -1160,7 +1171,7 @@ public function admin_notices() {
 				'<acronym title="' . esc_attr__( 'Content Delivery Network', 'w3-total-cache' ) . '">' . esc_html__( 'CDN', 'w3-total-cache' ) . '</acronym>'
 			),
 			'updated_pullzone_url'                    => __( 'Pull Zone URL could not be automatically updated. Please contact support for assistance.', 'w3-total-cache' ),
-			'cdn_totalcdn_fsd_origin_update_failed'  => __( 'Unable to update the Total CDN origin for Full Site Delivery. Please contact support for assistance.', 'w3-total-cache' ),
+			'cdn_totalcdn_fsd_origin_update_failed'   => __( 'Unable to update the Total CDN origin for Full Site Delivery. Please contact support for assistance.', 'w3-total-cache' ),
```
cssjoe: Use `W3TC_CDN_NAME` instead of "Total CDN".

Generic_Plugin_Admin.php
```diff
@@ -764,7 +764,15 @@ public function admin_print_scripts() {
 						)
 					),
 					'bunnyCdnWarning'  => esc_html__(
-						'Bunny CDN should only be enabled as either a CDN for static assets or full-site delivery, not both at the same time.  The CDN settings have been reverted.',
+						'Bunny CDN cannot be enabled for both CDN and Full Site Delivery.',
+						'w3-total-cache'
+					),
+					'totalCdnWarning'  => esc_html__(
+						'Total CDN cannot be enabled for both CDN and Full Site Delivery.',
```
cssjoe: Use `W3TC_CDN_NAME` instead of "Total CDN".

Generic_Plugin_Admin.php
```diff
@@ -764,7 +764,15 @@ public function admin_print_scripts() {
...
+						'w3-total-cache'
+					),
+					'mixedCdnWarning'  => esc_html__(
+						'Bunny CDN and Total CDN cannot be enabled across CDN and Full Site Delivery at the same time.',
```
cssjoe: Use `W3TC_CDN_NAME` instead of "Total CDN".

PR #1238, opened 2025-11-04
Fix: Add pagination handling to CloudFront distribution listing

ChadReitsma:
The CloudFront API will only return a **maximum** of 100 items per request. 

If a user has more than 100 distributions they cannot test or purge the CDN from within WP Admin if the distribution item is not returned within the first 100 results. 

These changes make the associated functions follow pagination markers to create an array of _all_ distributions (not just the first 100).


ChadReitsma:
Hey guys, is there something I can do to help speed up the merging process for this commit? It's a pretty simple fix.

ChadReitsma:
Hey guys, just a friendly push here. We've been waiting for 6 months for this to get integrated. Is this going to be possible?

PR #1261, opened 2025-12-01
Fix Util_File rmdir warnings

jacobd91:
This PR addresses potential warnings that are thrown by the Util_File.php rmdir method if malformed file paths are supplied to the method. Specifically, the provided path will now be sanitized before attempting file operations


mbtools:
Thank you! 👍

PR #1264, opened 2025-12-04
Update Setup Guide navigation handling

jacobd91:
https://imh-internal.atlassian.net/browse/ENG5-821

This PR does several things

1. It updates the "tabs" to be actual navigation tabs
2. It removes the "test required" restriction for the "Next" button, so you can freely go to the next or previous tab without running the test
3. It adds a checkmark to feature tabs on a visit to indicate that you have reviewed the content
4. Renamed "Image Optimization" tab to "Image Converter"

### DIFF
--- a/pub/js/setup-guide.js
+++ b/pub/js/setup-guide.js
@@ -9,6 +9,21 @@
  */
 
 var w3tc_enable_ga;
+var w3tcWizardSlideStepMap = {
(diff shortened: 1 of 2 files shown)


cssjoe (changes requested):
The checkmark should be applied after tests are completed and the user navigates to another slide.  For slides with no tests, the checkmark should be added when navigating to another slide.

PR #1265, opened 2025-12-04
Update Setup Guide tests to track test results and show average results

jacobd91:
https://imh-internal.atlassian.net/browse/ENG5-781

This PR does

1. Tracks test results and computes an average in a new column
2. Adds a new message instructing users to run tests multiple times and base their selection on the average results

### DIFF
--- a/pub/js/setup-guide.js
+++ b/pub/js/setup-guide.js
@@ -8,15 +8,12 @@
  * @global W3TC-setup-guide Localized array variable.
  */
 
-<<<<<<< HEAD
-var w3tc_enable_ga,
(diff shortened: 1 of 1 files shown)


cssjoe (changes requested):
1. The Page Cache slide text is too long and pushes the bottom of the box lower than the rest of the slides.  We should ensure that the box remains the same size for all slides.
2. I see the percentages next to the average times, but not the latest test times.  It would be useful to see them for both.
3. Review the Copilot comment.

cssjoe (approved):

PR #1282, opened 2026-01-09
Dunning and "Old Paypal" notices

jacobd91:
https://imh-internal.atlassian.net/browse/ENG5-878

This PR adds admin notices for

1. Subscription in dunning state (license state license.status)
2. Subscription on "Old Paypal" requiring re-authorization (license state license.paypal_billing_update_required)

### DIFF
--- a/Util_Ui.php
+++ b/Util_Ui.php
@@ -585,7 +585,7 @@ class Util_Ui {
 
 		if ( empty( $page_val ) || ( ! empty( $page_val ) && 'w3tc_' !== substr( $page_val, 0, 5 ) ) ) {
 			$logo = sprintf(
-				'<img src="%1s" alt="W3 Total Cache" style="height:30px;padding: 10px 2px 0 2px;" />"',
+				'<img src="%1$s" alt="W3 Total Cache" style="height:30px;padding: 10px 2px 0 2px;" />',
 				esc_url( plugins_url( '/pub/img/W3TC_dashboard_logo_title.png', W3TC_FILE ) ) . ''
 			);
 		} else {


Licensing_Plugin_Admin.php
```diff
@@ -283,122 +283,175 @@ private function _status_is( $s, $starts_with ) {
...
+						// Translators: 1 Total CDN tradmark, 2 opening HTML a tag to billing portal, 3 closing HTML a tag.
 						__(
-							'Your W3 Total Cache Pro license key has expired. %1$s to continue using the Pro features',
+							'Your %1$s Pro subscription payment is past due. Please update your %2$sBilling Information%3$s to prevent service interruption',
 							'w3-total-cache'
 						),
-						'<input type="button" class="button button-renew-plugin" data-nonce="' .
-							wp_create_nonce( 'w3tc' ) . '" data-renew-key="' . esc_attr( $this->get_license_key() ) .
-							'" data-src="licensing_expired" value="' . __( 'Renew Now', 'w3-total-cache' ) . '" />'
+						'Total Cache',
```
cssjoe: Use "W3 Total Cache".

Licensing_Plugin_Admin.php
```diff
@@ -638,6 +666,38 @@ private function get_license_notice( $status, $license_key ) {
...
+		$cached_url = get_transient( $transient_key );
+		if ( false !== $cached_url ) {
+			// Return cached value (empty string is a valid cached "no URL" result).
+			return $cached_url;
+		}
+
+		$billing_url = $this->fetch_billing_url_from_api( $license_key );
+
+		// Cache the result (including empty string for failed requests to avoid repeated failures).
+		set_transient( $transient_key, $billing_url, self::BILLING_URL_CACHE_DURATION );
```
cssjoe, line 768: I'm not sure why we would cache the billing URL only for an hour.  It's not likely to change.  Maybe increase that timeout to 24 hours or longer.

jacobd91: I actually dropped it down to 20 minutes as that's the lifetime of the token, after that point a new one would need to be generated. A cache was added so it doesn't make an API call on every page load if the notice is ignored

cssjoe (approved):

PR #1319, opened 2026-05-13
Block unserialize() object instantiation in cache stores

jacobd91:
## Summary                                                                                                               
   
  Closes the W3TC insecure-deserialization cluster (CHAIN-011, 3 findings). Every `unserialize()` call that reads          
  cache-store, transient, or extension-config payloads now passes `array( 'allowed_classes' => false )`, neutralizing the
  vendored `GuzzleHttp\Cookie\FileCookieJar::__destruct` POP-chain gadget at every documented entry point.                 
                                                            
  JIRA: **ENG7-3003**

  ## Vulnerability context

  | Bead | Severity | Description | Source |                                                                               
  |---|---|---|---|
  | rt9-39 | High | A08:2017 — `unserialize()` of AlwaysCached queue `extension` column |                                  `Extension_AlwaysCached_Plugin.php:173` |                                                                                
  | rt9-38 | High | A08:2017 — `unserialize()` of WP transient `w3tc_cdn_google_drive_folder_ids` (chains intoFileCookieJar `__destruct` → arbitrary file write) | `CdnEngine_GoogleDrive.php:640` |                                   
  | rt9-61 | High | `unserialize()` on cache-store payloads (Redis / Memcache / APC / WinCache / EAccelerator / File /Xcache / Apcu) | 14 sites across 7 cache backends |                                                                      
                                                            
  The vendored Guzzle `FileCookieJar::__destruct` calls `save($this->filename)` → `file_put_contents()` on an              
  attacker-controlled path, yielding an authenticated → unauthenticated RCE primitive when chained with any cache-write
  surface.                                                                                                                 
                                                            
  ## What changed

  17 call sites in 11 files now opt out of object instantiation during `unserialize()`:                                    
   
  | File | Lines |                                                                                                         
  |---|---|                                                 
  | `Cache_Apc.php` | 77, 144 |
  | `Cache_Apcu.php` | 77, 144 |
  | `Cache_Eaccelerator.php` | 77, 143 |                                                                                   
  | `Cache_File.php` | 187 |
  | `Cache_Redis.php` | 171, 409, 469 |                                                                                    
  | `Cache_Wincache.php` | 84, 156 |                                                                                       
  | `Cache_Xcache.php` | 87, 158 |
  | `Extension_AlwaysCached_Plugin.php` | 173 |                                                                            
  | `Cdn_Core_Admin.php` | 289 |                            
  | `CdnEngine_GoogleDrive.php` | 640 |                                                                                    
  | `Minify_MinifiedFileRequestHandler.php` | 1131 |                                                                       
   
  Before:                                                                                                                  
  ```php                                                    
  $v = @unserialize( apc_fetch( $storage_key ) );
  ```                                                                                                                      
  After:
  ```php                                                                                                                   
  $v = @unserialize( apc_fetch( $storage_key ), array( 'allowed_classes' => false ) );
  ```                                                                                                                      
   
  ## Why this approach (Option A: Layer 1 only)                                                                            
                                                            
  The alternative — migrating every backend's `set()`/`get()` pair from `serialize`/`unserialize` to JSON — is a flag-day  
  cache-shape change across 7 backends, high-risk for cache corruption on upgrade, and would orphan all existing cached
  entries. `allowed_classes => false` alone fully closes the RCE primitive at the read side because:                       
                                                            
  1. Every documented call site treats the unserialized value as an **array** (`is_array($v)` guard + array-key access), so
   the new `__PHP_Incomplete_Class` fallback for objects never affects normal operation — legitimate array payloads still
  deserialize as arrays.                                                                                                   
  2. `__PHP_Incomplete_Class` has no magic methods, so the Guzzle `FileCookieJar::__destruct` gadget (and any future
  class-instantiation gadget that lands in `vendor/`) cannot fire.                                                         
   
  JSON migration is recommended as a future hardening pass (defence-in-depth), not as a prerequisite for closing this CVE  
  class.                                                    
                                                                                                                           
  ## Files NOT changed (and why)                            

  - `vendor/guzzlehttp/guzzle/src/Cookie/FileCookieJar.php` — vendored, the gadget is unreachable once Layer 1 lands.      
  - `Cache_Base.php` — already carries the `phpcs:disable serialize_unserialize` pragma but contains no active
  `unserialize()` call.                                                                                                    
  - `lib/Google/Cache/File.php`, `lib/Minify/Minify/Cache/File.php`, `lib/Nusoap/class.wsdlcache.php` — third-party `lib/`
  directory, excluded from `phpcs.xml` and out of audit scope.                                                             
                                                            
  ## Caller compatibility audit                                                                                            
                                                            
  | Site | Caller-side guard |                                                                                             
  |---|---|
  | `Cache_*::get_with_old` | `is_array($v) && isset($v['key_version'])` |                                                 
  | `Cache_*::delete` | `is_array($v)` then `$v['key_version'] = 0` |
  | `Cdn_Core_Admin.php:289` | iterates `$metadata->files` array via existing accessor |                                   
  | `CdnEngine_GoogleDrive.php:640` | `isset($ids[$key])` array access only |                                              
  | `Extension_AlwaysCached_Plugin.php:173` | result passed through `w3tc_pagecache_set` filter, treated opaquely |        
  | `Minify_MinifiedFileRequestHandler.php:1131` | result keyed as `$value['content']` etc. |                              
                                                                                                                           
  No caller invokes a method on the unserialized result — `allowed_classes => false` is behaviourally a no-op for          
  legitimate array-shaped cache entries.                                                                                   
                                                                                                                           
  ## Automated test plan                                    

  Run from the plugin root:

  ``` 
  bash
  1. PHP syntax check on all modified files
  php -l Cache_Apc.php Cache_Apcu.php Cache_Eaccelerator.php Cache_File.php \                                              
         Cache_Redis.php Cache_Wincache.php Cache_Xcache.php \                                                             
         CdnEngine_GoogleDrive.php Cdn_Core_Admin.php \                                                                    
         Extension_AlwaysCached_Plugin.php Minify_MinifiedFileRequestHandler.php                                           
                                                            
  2. Lint everything (catches any wider regression)                                                                      
  yarn run php-lint                                                                                                        
   
  3. WordPress coding-standards check on the diff                                                                        
  ./vendor/bin/phpcs Cache_Apc.php Cache_Apcu.php Cache_Eaccelerator.php \                                                 
                     Cache_File.php Cache_Redis.php Cache_Wincache.php \                                                   
                     Cache_Xcache.php CdnEngine_GoogleDrive.php \                                                          
                     Cdn_Core_Admin.php Extension_AlwaysCached_Plugin.php \                                                
                     Minify_MinifiedFileRequestHandler.php  
                                                                                                                           
  4. PHPUnit (requires WP test env — see CLAUDE.md "Running tests")
  ./vendor/bin/phpunit                                                                                                     
                                                            
  5. Static unserialize call-site audit — must return zero un-flagged sites                                              
  grep -nE '@?unserialize\(' Cache_*.php Cdn_Core_Admin.php \                                                              
       CdnEngine_GoogleDrive.php Extension_AlwaysCached_Plugin.php \                                                       
       Minify_MinifiedFileRequestHandler.php \                                                                             
     | grep -v "allowed_classes"                                                                                           
  Expected output: empty                                                                                                 
  ```                                                       
                                                                                                                           
  Expected: all lint clean, all unserialize call sites carry `allowed_classes => false`, no regression in PHPUnit suite.   
   
  ## Manual test plan                                                                                                      
                                                            
  Test on a fresh WordPress install with W3TC active. Repeat for each backend the host environment supports.               
   
  A. Functional regression — caches still work                                                                         
                                                            
  For each engine the host supports (`File`, `Redis`, `Memcache`, `APC`/`APCu`, `WinCache`, `Xcache`, `EAccelerator`):     
   
  1. **Page Cache:** *Performance → Page Cache → Method = $engine → Save*. Load a page logged-out twice; confirm second hit
   is served from cache (`X-W3TC-PageCache` header or `<!-- Page Cache -->` footer marker).
  2. **Object Cache:** *Performance → Object Cache → Method = $engine → Save*. Confirm dashboard widgets, admin transients,
   and `get_option()` reads continue to return correct values across requests.                                             
  3. **Database Cache:** *Performance → Database Cache → Method = $engine → Save*. Confirm post lists, search results, and
  front-end queries render normally.                                                                                       
  4. **Empty all caches** via *Performance → Dashboard → Empty all caches*. Confirm no PHP warnings/notices in
  `wp-content/debug.log` (enable `WP_DEBUG_LOG`).                                                                          
                                                            
  B. CDN GoogleDrive flow                                                                                              
                                                            
  The transient `w3tc_cdn_google_drive_folder_ids` is read at `CdnEngine_GoogleDrive.php:640`. If a GoogleDrive CDN is     
  configured:
                                                                                                                           
  1. *Performance → CDN → Configure → push a small file via "Manual: Upload changes"*.                                     
  2. Confirm the folder mapping read still works (no `Undefined index` or fatal in `debug.log`).
  3. If GoogleDrive isn't in your test env, exercise the read path via WP-CLI:                                             
     ```bash                                                                                                               
     wp transient set w3tc_cdn_google_drive_folder_ids 'a:1:{s:4:"test";s:6:"abc123";}'                                    
     wp eval 'var_dump( get_transient("w3tc_cdn_google_drive_folder_ids") );'                                              
     # expect: array(1) { ["test"]=> string(6) "abc123" }                                                                  
     ```                                                                                                                   
                                                                                                                           
  C. AlwaysCached extension flow                        
                                                                                                                           
  If AlwaysCached extension is active:                      

  1. *Performance → Extensions → AlwaysCached → Activate*.                                                                 
  2. Add a URL to the queue via the extension UI.
  3. Trigger a cache warm; confirm the queue processes without fatal/warning in `debug.log` (the read site is              
  `Extension_AlwaysCached_Plugin.php:173`).                                                                                
   
  D. Security regression — POP-chain neutralized                                                                       
                                                            
  This is the load-bearing manual test. Confirm a crafted serialized object **does not** instantiate.                      
   
  ```bash                                                                                                                  
  Inject a fake FileCookieJar payload as a WP transient   
  wp transient set w3tc_cdn_google_drive_folder_ids \                                                                      
    'O:33:"GuzzleHttp\Cookie\FileCookieJar":2:{s:8:"filename";s:30:"/tmp/w3tc_deser_proof_BAD.json";s:13:"storeSessionCooki
  es";b:1;}'                                                                                                               
                                                            
  Force the read path + destructor trigger                                                                               
  wp eval '                                                                                                                
    $v = get_transient("w3tc_cdn_google_drive_folder_ids");
    $v2 = @unserialize($v, array("allowed_classes" => false));                                                             
    var_dump($v2);                                                                                                         
    unset($v2);
  '                                                                                                                        
                                                            
  Expected:
     var_dump shows object(__PHP_Incomplete_Class) — not GuzzleHttp\Cookie\FileCookieJar
     /tmp/w3tc_deser_proof_BAD.json DOES NOT EXIST after the eval finishes                                                
  test ! -f /tmp/w3tc_deser_proof_BAD.json && echo "PASS: no file written"                                                 
                                                                                                                           
  Cleanup                                                                                                                
  wp transient delete w3tc_cdn_google_drive_folder_ids                                                                     
  ```                                                                                                                      
   
  Pre-patch behaviour (for comparison, do **not** run on production): `unserialize()` would instantiate `FileCookieJar`,   
  `unset()` would trigger `__destruct → save() → file_put_contents()`, and `/tmp/w3tc_deser_proof_BAD.json` would exist.
  Post-patch: the unserialize returns `__PHP_Incomplete_Class`, no destructor fires, no file is written.                   
                                                            
  E. Existing cache contents on upgrade

  Sites upgrading from 2.9.4 may have cache entries serialized with no class restriction. Confirm:                         
   
  1. Pre-populate the page cache **on the pre-patch build** with normal traffic.                                           
  2. Deploy the patched build.                              
  3. Reload pages — they should still serve from cache (the stored payloads are array-shaped, which `allowed_classes =>    
  false` happily returns).                                                                                                 
  4. No PHP notices / warnings in `debug.log`.
                                                                                                                           
  ## Rollback                                               

  Trivial revert — every change is the same single-arg → two-arg call signature. No new files, no schema migration, no     
  config changes. `git revert <commit>` restores prior behaviour exactly.
                                                                                                                           
  ## Verification artifacts                                                                                                
   
  - Static verification report: `tmp/sec-verify/deserialization-static-report.md`                                          
  - Progress tracker entry: `tmp/sec-progress.json` → `deserialization` (`status: verified-static`, 3/3 matches-pattern)
                                                                                                                           
  ## Out of scope / follow-up                                                                                              
                                                                                                                           
  - **JSON-write migration** for the 7 cache backends — defence-in-depth, future hardening pass.                           
  - **HMAC-sign cache payloads** (same pattern as the `eval-rce` group's sign-at-write) — would also stop array-shape
  injection in backends with attacker-shared write surface.                                                                
  - Vendored Guzzle `FileCookieJar` is left intact (the gadget is unreachable; rewriting vendored code is out of scope).

### DIFF
--- a/Cache_File.php
+++ b/Cache_File.php
@@ -184,7 +184,7 @@ class Cache_File extends Cache_Base {
 	public function get_with_old( $key, $group = '' ) {
 		list( $data, $has_old_data ) = $this->_get_with_old_raw( $key, $group );
 		if ( ! empty( $data ) ) {
-			$data_unserialized = @unserialize( $data );
+			$data_unserialized = @unserialize( $data, array( 'allowed_classes' => false ) );
 		} else {
 			$data_unserialized = $data;
 		}
--- a/Cache_Redis.php
+++ b/Cache_Redis.php
@@ -466,7 +466,13 @@ class Cache_Redis extends Cache_Base {
 
 			foreach ( $orig_keys as $i => $orig_key ) {
 				if ( isset( $values[ $i ] ) && false !== $values[ $i ] ) {
-					$results[ $orig_key ] = @unserialize( $values[ $i ], array( 'allowed_classes' => false ) );
+					$decoded = @unserialize( $values[ $i ], array( 'allowed_classes' => false ) ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+
+					// `allowed_classes => false` returns `__PHP_Incomplete_Class`
+					// for crafted object payloads. Callers of get_multi() expect
+					// each value to be either null (miss) or an array (legitimate
+					// cache entry), so coerce non-arrays to null.
+					$results[ $orig_key ] = is_array( $decoded ) ? $decoded : null;
 				} else {
 					$results[ $orig_key ] = null;
 				}
--- a/Cache_Redis.php
+++ b/Cache_Redis.php
@@ -168,7 +168,7 @@ class Cache_Redis extends Cache_Base {
 		}
 
 		$v = $accessor->get( $storage_key );
-		$v = @unserialize( $v );
+		$v = @unserialize( $v, array( 'allowed_classes' => false ) );
 
 		if ( ! is_array( $v ) || ! isset( $v['key_version'] ) ) {
 			return array( null, $has_old_data );
@@ -406,7 +406,7 @@ class Cache_Redis extends Cache_Base {
 		$accessor->watch( $storage_key );
 
 		$value = $accessor->get( $storage_key );
-		$value = @unserialize( $value );
+		$value = @unserialize( $value, array( 'allowed_classes' => false ) );
 
 		if ( ! is_array( $value ) ) {
 			$accessor->unwatch();
@@ -466,7 +466,7 @@ class Cache_Redis extends Cache_Base {
 
 			foreach ( $orig_keys as $i => $orig_key ) {
 				if ( isset( $values[ $i ] ) && false !== $values[ $i ] ) {
-					$results[ $orig_key ] = @unserialize( $values[ $i ] );
+					$results[ $orig_key ] = @unserialize( $values[ $i ], array( 'allowed_classes' => false ) );
 				} else {
 					$results[ $orig_key ] = null;
 				}
--- a/CdnEngine_GoogleDrive.php
+++ b/CdnEngine_GoogleDrive.php
@@ -637,7 +637,7 @@ class CdnEngine_GoogleDrive extends CdnEngine_Base {
 	private function parent_id_resolve_step( $root_id, $folder ) {
 		// decode top folder.
 		$ids_string = get_transient( 'w3tc_cdn_google_drive_folder_ids' );
-		$ids        = @unserialize( $ids_string );
+		$ids        = @unserialize( $ids_string, array( 'allowed_classes' => false ) );
 
 		if ( isset( $ids[ $root_id . '_' . $folder ] ) ) {
 			return $ids[ $root_id . '_' . $folder ];
--- a/CdnEngine_GoogleDrive.php
+++ b/CdnEngine_GoogleDrive.php
@@ -637,7 +637,14 @@ class CdnEngine_GoogleDrive extends CdnEngine_Base {
 	private function parent_id_resolve_step( $root_id, $folder ) {
 		// decode top folder.
 		$ids_string = get_transient( 'w3tc_cdn_google_drive_folder_ids' );
-		$ids        = @unserialize( $ids_string, array( 'allowed_classes' => false ) );
+		$ids        = @unserialize( $ids_string, array( 'allowed_classes' => false ) ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+
+		// `allowed_classes => false` substitutes `__PHP_Incomplete_Class` for any
+		// serialized object payload; array-offset access on that would fatal.
+		// Normalize to an empty array up-front so every later access is safe.
+		if ( ! is_array( $ids ) ) {
+			$ids = array();
+		}
 
 		if ( isset( $ids[ $root_id . '_' . $folder ] ) ) {
 			return $ids[ $root_id . '_' . $folder ];
@@ -676,10 +683,7 @@ class CdnEngine_GoogleDrive extends CdnEngine_Base {
 			$this->_service->permissions->insert( $id, $permission );
 		}
 
-		if ( ! is_array( $ids ) ) {
-			$ids = array();
-		}
-
+		// `$ids` was normalized to an array immediately after unserialize() above.
 		$ids[ $root_id . '_' . $folder ] = $id;
 		set_transient( 'w3tc_cdn_google_drive_folder_ids', serialize( $ids ) );
 
--- a/Cache_Apc.php
+++ b/Cache_Apc.php
@@ -74,7 +74,7 @@ class Cache_Apc extends Cache_Base {
 		$has_old_data = false;
 		$storage_key  = $this->get_item_key( $key );
 
-		$v = @unserialize( apc_fetch( $storage_key ) );
+		$v = @unserialize( apc_fetch( $storage_key ), array( 'allowed_classes' => false ) );
 		if ( ! is_array( $v ) || ! isset( $v['key_version'] ) ) {
 			return array( null, $has_old_data );
 		}
@@ -141,7 +141,7 @@ class Cache_Apc extends Cache_Base {
 		$storage_key = $this->get_item_key( $key );
 
 		if ( $this->_use_expired_data ) {
-			$v = @unserialize( apc_fetch( $storage_key ) );
+			$v = @unserialize( apc_fetch( $storage_key ), array( 'allowed_classes' => false ) );
 			if ( is_array( $v ) ) {
 				$v['key_version'] = 0;
 				apc_store( $storage_key, serialize( $v ), 0 );
--- a/Cache_Apcu.php
+++ b/Cache_Apcu.php
@@ -74,7 +74,7 @@ class Cache_Apcu extends Cache_Base {
 		$has_old_data = false;
 		$storage_key  = $this->get_item_key( $key );
 
-		$v = @unserialize( apcu_fetch( $storage_key ) );
+		$v = @unserialize( apcu_fetch( $storage_key ), array( 'allowed_classes' => false ) );
 		if ( ! is_array( $v ) || ! isset( $v['key_version'] ) ) {
 			return array( null, $has_old_data );
 		}
@@ -141,7 +141,7 @@ class Cache_Apcu extends Cache_Base {
 		$storage_key = $this->get_item_key( $key );
 
 		if ( $this->_use_expired_data ) {
-			$v = @unserialize( apcu_fetch( $storage_key ) );
+			$v = @unserialize( apcu_fetch( $storage_key ), array( 'allowed_classes' => false ) );
 			if ( is_array( $v ) ) {
 				$v['key_version'] = 0;
 				apcu_store( $storage_key, serialize( $v ), 0 );
--- a/Cache_Eaccelerator.php
+++ b/Cache_Eaccelerator.php
@@ -74,7 +74,7 @@ class Cache_Eaccelerator extends Cache_Base {
 		$has_old_data = false;
 		$storage_key  = $this->get_item_key( $key );
 
-		$v = @unserialize( eaccelerator_get( $storage_key ) );
+		$v = @unserialize( eaccelerator_get( $storage_key ), array( 'allowed_classes' => false ) );
 		if ( ! is_array( $v ) || ! isset( $v['key_version'] ) ) {
 			return array( null, $has_old_data );
 		}
@@ -140,7 +140,7 @@ class Cache_Eaccelerator extends Cache_Base {
 		$storage_key = $this->get_item_key( $key );
 
 		if ( $this->_use_expired_data ) {
-			$v = @unserialize( eaccelerator_get( $storage_key ) );
+			$v = @unserialize( eaccelerator_get( $storage_key ), array( 'allowed_classes' => false ) );
 			if ( is_array( $v ) ) {
 				$v['key_version'] = 0;
 				eaccelerator_put( $storage_key, serialize( $v ), 0 );
--- a/Cache_Wincache.php
+++ b/Cache_Wincache.php
@@ -81,7 +81,7 @@ class Cache_Wincache extends Cache_Base {
 
 		$storage_key = $this->get_item_key( $key );
 
-		$v = @unserialize( wincache_ucache_get( $storage_key ) );
+		$v = @unserialize( wincache_ucache_get( $storage_key ), array( 'allowed_classes' => false ) );
 		if ( ! is_array( $v ) || ! isset( $v['key_version'] ) ) {
 			return array( null, $has_old_data );
 		}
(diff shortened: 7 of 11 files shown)


cssjoe (commented):
## Verdict

LGTM with one optional follow-up nit. Tightly scoped, well-documented security fix that closes the CHAIN-011 deserialization cluster correctly. The two-commit shape (Layer 1: `allowed_classes => false` everywhere; Layer 2: caller-side normalization where needed) is exactly the right way to address the Copilot feedback without thrashing the patch.

## Scope check

`grep -nE '@?unserialize\(' Cache_*.php Cdn_Core_Admin.php CdnEngine_GoogleDrive.php Extension_AlwaysCached_Plugin.php Minify_MinifiedFileRequestHandler.php | grep -v allowed_classes` returns only one match (a comment line in `CdnEngine_GoogleDrive.php`) — every active `unserialize()` call in the audited cluster now opts out of class instantiation.

In-tree `unserialize()` calls *not* in this PR — confirmed correctly excluded:
- `lib/Google/Cache/File.php`, `lib/Minify/Minify/Cache/File.php`, `lib/Nusoap/class.wsdlcache.php` — `lib/` is `phpcs.xml`-excluded vendored third-party.
- `qa/plugins/fragmentcache/start-end.php` — QA test fixture.
- `Cache_Memcache.php`, `Cache_Memcached.php`, `Cache_Nginx_Memcached.php` — confirmed no `unserialize` calls; the extensions deserialize natively in C.
- `Cache_Base.php` — has the file-level `phpcs:disable serialize_unserialize` pragma but no active `unserialize` call.

## Layer-2 normalization audit (the part Copilot flagged)

Verified the PHP 8 `__PHP_Incomplete_Class` array-access behavior the second commit message asserts:

```text
$ php -r 'class Foo{public $bar="baz";} $v=unserialize(serialize(new Foo()),["allowed_classes"=>false]); var_dump(isset($v["bar"]));'
ERROR: Cannot use object of type __PHP_Incomplete_Class as array
```

So the motivation for the second commit's `is_array()`/`is_object()` coercions is real — without them the RCE primitive would have been swapped for a remote-DoS primitive. Walked each Layer-2 site against its caller:

| Site | Existing guard? | Layer-2 fix | Verdict |
|---|---|---|---|
| `Cache_Apc/Apcu/Eaccelerator/Wincache/Xcache::get_with_old` | `! is_array($v) \|\| ! isset(...)` short-circuits | not needed | OK |
| `Cache_Apc/Apcu/Eaccelerator/Wincache/Xcache::delete` | `if ( is_array( $v ) )` | not needed | OK |
| `Cache_Redis.php:171` | `! is_array($v) \|\| ! isset(...)` | not needed | OK |
| `Cache_Redis.php:409` (`set_if_maybe_equals`) | `if ( ! is_array( $value ) )` | not needed | OK |
| `Cache_Redis.php:469` (`get_multi`) | none | non-array → `null` | OK, matches caller in `ObjectCache_WpObjectCache_Regular.php:360` which checks `is_array($v)` |
| `CdnEngine_GoogleDrive.php:640` | later `is_array` block, run after side-effects | hoisted; later block removed | OK, no intermediate mutation of `$ids` between lines 645 and 687 |
| `Cdn_Core_Admin.php:289` | none | non-array → `array()` | OK, `Cdn_Core::get_metadata_files()` uses `isset($metadata['file'])` |
| `Extension_AlwaysCached_Plugin.php:173` | none | non-array → `array()` | OK, `w3tc_pagecache_set()` does offset access |
| `Minify_MinifiedFileRequestHandler.php:1131` | none | object → `false` | OK, both callers (lines 357, 519) treat `false` as a miss |

## Local verification ran on the PR branch

- `yarn run php-lint` — clean.
- `./vendor/bin/phpcs <11 changed files>` — clean.
- `php tests/test-mfunc-security.php` — 28/28 pass (unrelated cluster, sanity).
- Static audit from the PR's automated test plan — clean.

## Minor nits (non-blocking)

1. PR description's rt9-61 row mentions `Memcache`, but `Cache_Memcache.php` / `Cache_Memcached.php` correctly weren't touched (no `unserialize` call — extensions deserialize natively). A one-line note in the body would save future auditors a re-flag.
2. Codecov reports 0% patch coverage. Acknowledged via the manual `FileCookieJar` regression test in section D, which is the load-bearing verification. A PHPUnit fixture round-tripping a serialized object through `Cache_File::get_with_old` would lock in the behavior for future refactors — out of scope here.
3. See inline notes on `Cache_File.php:187` (optional DoS surface left open) and `Cache_Redis.php:434` (docblock now slightly inaccurate).

Approve pending @cssjoe's formal review.

Cache_File.php
```diff
@@ -184,7 +184,7 @@ public function set( $key, $content, $expiration = 0, $group = '' ) {
 	public function get_with_old( $key, $group = '' ) {
 		list( $data, $has_old_data ) = $this->_get_with_old_raw( $key, $group );
 		if ( ! empty( $data ) ) {
-			$data_unserialized = @unserialize( $data );
+			$data_unserialized = @unserialize( $data, array( 'allowed_classes' => false ) );
```
cssjoe, line 187: Optional follow-up — `get_with_old` is the only Layer-2 read site without an `is_array` coercion. If a `Cache_File`-backed page cache entry is poisoned with a serialized object, `$data_unserialized` becomes `__PHP_Incomplete_Class` and the immediate caller `PgCache_ContentGrabber::_extract_cached_page` (`$data['compression'] = $compression;` ~line 391) will fatal in PHP 8.x.

Not a regression — pre-patch the same caller would have fataled on any serialized scalar already — and exploiting it requires write access under the cache dir (a much weaker threat model than the transient/postmeta/DB surfaces this PR targets), so fine to merge as-is. The object-cache path that also goes through `Cache_File` is safe because `ObjectCache_WpObjectCache_Regular::get` (line 228) checks `is_array($v)` itself.

For symmetry with the other five Layer-2 sites, a one-line coercion would be nice:

```suggestion
			$data_unserialized = @unserialize( $data, array( 'allowed_classes' => false ) );
			if ( ! is_array( $data_unserialized ) ) {
				$data_unserialized = null;
			}
```

Alternatively, a comment noting that the page-cache file path relies on the cache dir being write-protected would be enough.

Cache_Redis.php
```diff
@@ -466,7 +466,13 @@ public function get_multi( array $keys, $group = '' ) {
 
 			foreach ( $orig_keys as $i => $orig_key ) {
 				if ( isset( $values[ $i ] ) && false !== $values[ $i ] ) {
-					$results[ $orig_key ] = @unserialize( $values[ $i ] );
+					$decoded = @unserialize( $values[ $i ], array( 'allowed_classes' => false ) ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
```
cssjoe, line 469: Trivial nit (pre-existing, but a bit more visible after this PR): the `get_multi` docblock at line 434 reads `@return array Map of cache key => raw cached payload (serialized array or null).` The value is deserialized, not raw, and after this change is explicitly coerced to `null` on object payloads. A future-pass docblock update could read:

```
@return array Map of cache key => deserialized array payload, or null for misses / invalid entries.
```

No need to fix in this PR.

CdnEngine_GoogleDrive.php
```diff
@@ -676,10 +683,7 @@ private function parent_id_resolve_step( $root_id, $folder ) {
 			$this->_service->permissions->insert( $id, $permission );
 		}
 
-		if ( ! is_array( $ids ) ) {
-			$ids = array();
-		}
-
+		// `$ids` was normalized to an array immediately after unserialize() above.
```
cssjoe, line 686: Verified safe — between the new early `if ( ! is_array( $ids ) ) { $ids = array(); }` on line 645 and this line, no code path mutates `$ids` (the only reads are `isset($ids[...])` and the `return $ids[...]` on a hit). The hoist + comment is the right cleanup.

cssjoe (approved):

PR #1320, opened 2026-05-13
ENG7-3007

jacobd91:
https://imh-internal.atlassian.net/browse/ENG7-3007

### DIFF
--- a/ConfigKeysSchema.php
+++ b/ConfigKeysSchema.php
@@ -238,13 +238,20 @@ class ConfigKeysSchema {
 		// `__toString`, so an attacker-shaped object payload would
 		// otherwise become a fatal instead of a silent drop.
 		$is_object = \is_object( $value );
+		$is_array  = \is_array( $value );
 
 		switch ( $descriptor['type'] ) {
 			case 'boolean':
-				return $is_object ? false : (bool) $value;
+				// Boolean keys accept only scalars. `(bool) ['x'=>1]`
+				// would otherwise be `true`, smuggling a structured
+				// payload through a toggle.
+				if ( $is_object || $is_array ) {
+					return false;
+				}
+				return (bool) $value;
 
 			case 'integer':
-				if ( $is_object || \is_array( $value ) ) {
+				if ( $is_object || $is_array ) {
 					return 0;
 				}
 				return (int) $value;
@@ -253,7 +260,7 @@ class ConfigKeysSchema {
 				return \is_scalar( $value ) ? (string) $value : '';
 
 			case 'array':
-				return \is_array( $value ) ? $value : array();
+				return $is_array ? $value : array();
 		}
 
 		return $value;
--- a/ConfigKeysSchema.php
+++ b/ConfigKeysSchema.php
@@ -1,0 +1,293 @@
+<?php
+/**
+ * File: ConfigKeysSchema.php
+ *
+ * @package W3TC
+ *
+ * @since X.X.X
+ */
+
+namespace W3TC;
+
+/**
+ * Class ConfigKeysSchema
+ *
+ * Read-only accessor for the configuration-key schema in `ConfigKeys.php`.
+ *
+ * The schema file itself is a flat `$keys` array — every legitimate W3TC
+ * configuration key, with its declared type, default value, and (since the
+ * mass-assignment hardening pass) optional `flags` map. Treat the file as
+ * the source of truth: any inbound write that originates from user input
+ * MUST validate the key against this schema before reaching `Config::set()`.
+ *
+ * This class exposes:
+ *  - `get_keys()`      — the full schema array (loaded once and cached).
+ *  - `is_known()`      — whether a key is in the schema.
+ *  - `descriptor()`    — the per-key descriptor (or null).
+ *  - `can_import()`    — whether a key may be written by bulk-import flows
+ *                        (Config::import). High-impact keys (java executable
+ *                        paths, the active-extensions array, cache-engine
+ *                        switches) are marked `no_import => true` and are
+ *                        rejected here even though they are otherwise
+ *                        legitimate keys.
+ *  - `coerce()`        — type-coerce a raw value to its declared type
+ *                        (boolean/integer/string/array). Anything outside
+ *                        the declared type is reduced to the type's safe
+ *                        default; this stops object payloads (POP gadgets)
+ *                        and PHP-source strings from landing in toggles
+ *                        and counts.
+ *
+ * `Config::set()` is intentionally NOT gated by this schema. Internal
+ * callers (Mobile_Base, Extensions_Util, etc.) write keys computed at
+ * runtime from per-engine slugs that aren't reflected in the static
+ * schema; making `set()` strict would break extension activation. The
+ * correct gate sits at the user-input boundary — `Config::import()`,
+ * `read_request()`, the dbcluster-save handler, and the overloaded
+ * toggle endpoints — where the trust changes.
+ *
+ * @since X.X.X
+ */
+class ConfigKeysSchema {
+	/**
+	 * Cached schema array. Loaded lazily from `ConfigKeys.php` on first
+	 * access and reused for the rest of the request.
+	 *
+	 * @var array<string,array>|null
+	 */
+	private static $cache = null;
+
+	/**
+	 * Set of state keys writable by `w3tc_default_config_state` /
+	 * `w3tc_default_config_state_master`. These are persisted into the
+	 * `w3tc_state` option (not the main W3TC config) and are written by
+	 * dismissable-notice / wizard handlers that previously accepted any
+	 * key name from `$_GET`. Lock the set to known operational toggles.
+	 *
+	 * @var string[]
+	 */
+	private static $state_key_allowlist = array(
+		'common.hide_note_no_general_settings_modified',
+		'common.hide_note_php_is_old',
+		'common.show_note_setup_guide_on_settings_save',
+		'common.support_us',
+		'license.community_terms',
+		'license.next_check',
+		'license.status',
+		'license.terms',
+		'minify.show_note.need_flush',
+		'minify.show_note.need_empty_cache_dir',
+		'objectcache.show_note.flush_needed',
+		'pagecache.show_note.flush_needed',
+		'common.install',
+		'common.install_version',
+		'license.has_pro_features',
+		'license.expires',
+		'extension.swarmify.hide_note_swarmify_active',
+		'extension.amp.hide_note_amp_active',
+		'common.show_note.activated',
+		'common.show_note.cleanup_needed',
+		'common.show_note.notes_disabled',
+		'license.expiration_date',
+		'license.last_check',
+	);
+
+	/**
+	 * Returns the full configuration-key schema.
+	 *
+	 * Loaded from `ConfigKeys.php` (which defines a flat `$keys` array)
+	 * once per request. The result is identical to the legacy
+	 * `include W3TC_DIR . '/ConfigKeys.php'; // defines $keys` pattern
+	 * used in `read_request()` and elsewhere — callers can replace
+	 * scattered `include`s with a single call here.
+	 *
+	 * @since X.X.X
+	 *
+	 * @return array<string,array>
+	 */
+	public static function get_keys() {
+		if ( null === self::$cache ) {
+			$keys = array();
+			include __DIR__ . '/ConfigKeys.php';
+			self::$cache = \is_array( $keys ) ? $keys : array();
+		}
+
+		return self::$cache;
+	}
+
+	/**
+	 * Returns true if the given key is documented in the schema.
+	 *
+	 * Compound keys (extension subkeys, expressed as a `[parent, child]`
+	 * tuple) are accepted unconditionally because the parent slot in the
+	 * schema is `extensions.settings` and the child slot is owned by the
+	 * extension. The gate for those lives in the extension's own handler.
+	 *
+	 * @since X.X.X
+	 *
+	 * @param string|array $key Single-string or [parent, child] compound key.
+	 *
+	 * @return bool
+	 */
+	public static function is_known( $key ) {
+		if ( \is_array( $key ) ) {
+			return true;
+		}
+
+		if ( ! \is_string( $key ) || '' === $key ) {
+			return false;
+		}
+
+		$keys = self::get_keys();
+
+		return \array_key_exists( $key, $keys );
+	}
+
+	/**
+	 * Returns the descriptor for a key, or null if not present.
+	 *
+	 * @since X.X.X
+	 *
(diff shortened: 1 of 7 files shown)


Generic_AdminActions_Config.php
```diff
@@ -201,6 +236,152 @@ public function w3tc_config_dbcluster_config_save() {
...
+			if ( \in_array( $tid, $forbidden_tokens, true ) ) {
+				return \sprintf(
+					/* translators: %s: token name. */
+					\__( 'Disallowed PHP construct in configuration file: %s', 'w3-total-cache' ),
+					\token_name( $tid )
+				);
+			}
+
+			// A bareword followed by `(` is a function call.
+			if ( T_STRING === $tid ) {
```
cssjoe, line 337: **Blocking — confirmed RCE bypass.** The "`T_STRING` followed by `(`" check is the only thing distinguishing function calls from `array(`, but several other token shapes also accept `(` and aren't covered. On PHP 8.3.31 I confirmed (validator function lifted into a standalone probe) that all six of these payloads ACCEPT, and require'ing the resulting file executes `system('id')` end-to-end:

```php
// 1. variable function — T_VARIABLE followed by `(` is not checked.
$w3tc_dbcluster_config = array();
$cmd = 'system';
$cmd('id');

// 2. namespaced call on PHP 8+ — `\system` is a single T_NAME_FULLY_QUALIFIED
//    token, not T_STRING, so the bareword check never fires.
$w3tc_dbcluster_config = array();
\system('id');

// 3. qualified namespace — same story, T_NAME_QUALIFIED.
NS\bad('id');

// 4. arrow function + variable invoke — T_FN is not in $forbidden_tokens
//    (line 282 above), and the body uses the variable-function path.
$evil = fn($x) => $x('id');
$evil('system');

// 5. string literal as callable — `("system")("id")` is `)` followed by `(`.
("system")("id");

// 6. parenthesized assignment as callable.
($x = 'system')('id');
```

W3TC supports PHP 7.2.5–8.3 (per `composer.json`), so the PHP 8 name-token shapes are in scope.

**Recommended fix:** invert to an allowlist for `(` — for every `(` token, walk back past `T_WHITESPACE`/`T_COMMENT`/`T_DOC_COMMENT` and reject unless the previous significant token is exactly `T_STRING` with value `array`. One rule covers every call shape (variable, namespaced, qualified, string-literal, parenthesized) and is robust against future PHP versions adding new token types. Also add `T_FN` to `$forbidden_tokens` at line 282 — the dbcluster file should never declare callables.

The six payloads above should be added to `test_validate_rejects_code_execution_primitives()` in `class-w3tc-dbcluster-validator-test.php`.

Generic_AdminActions_Config.php
```diff
@@ -201,6 +236,152 @@ public function w3tc_config_dbcluster_config_save() {
...
+	 * @param string $content Raw submitted PHP source.
+	 *
+	 * @return string|null Localised error string on rejection, or null on accept.
+	 */
+	public static function validate_dbcluster_content( $content ) {
+		if ( ! \is_string( $content ) || '' === \trim( $content ) ) {
+			return \__( 'Configuration file body is empty.', 'w3-total-cache' );
+		}
+
+		if ( 0 !== \strpos( \ltrim( $content ), '<?php' ) ) {
```
cssjoe, line 264: Nit: PHP accepts `<?php`, `<?PHP`, `<?Php` as the open tag, but this check is case-sensitive. A user pasting a tool-generated file with mixed-case open tag will see "must start with `<?php`" and have no idea their file is valid PHP. Either accept case-insensitively with `stripos(...) === 0` or update the message to spell out "lowercase `<?php`".

ConfigKeysSchema.php
```diff
@@ -0,0 +1,300 @@
...
+		'license.has_pro_features',
+		'license.expires',
+		'extension.swarmify.hide_note_swarmify_active',
+		'extension.amp.hide_note_amp_active',
+		'common.show_note.activated',
+		'common.show_note.cleanup_needed',
+		'common.show_note.notes_disabled',
+		'license.expiration_date',
+		'license.last_check',
+	);
```
cssjoe, line 92: **Blocking — functional regression.** This list is missing every dismissable-notice key actually used in the plugin. A grep across the codebase for `'w3tc_default_config_state{,_master,_note}' => 'y'` paired with the `'key' => '...'` next to it yields 21 keys, **zero of which** are in this allowlist:

```
cdn.hide_note_no_curl                        common.show_note.flush_posts_needed
cdn.show_note_cdn_reupload                   common.show_note.flush_statics_needed
cdn.show_note_cdn_upload                     common.show_note.nginx_restart_required
cdn.show_note_theme_changed                  common.show_note.plugins_updated
cdn.show_note_wp_upgraded                    genesis.theme.hide_note_suggest_activation
common.hide_note_debug_enabled               minify.show_note_minify_error
common.hide_note_no_permalink_rules          newrelic.hide_note_pageload_slow
common.hide_note_no_zlib                     wordpress_seo.hide_note_suggest_activation
common.hide_note_stats_enabled               wpml.hide_note_language_negotiation_type
common.hide_note_wp_content_permissions      wpml.hide_note_suggest_activation
common.hide_note_zlib_output_compression
```

These are the keys behind the "X" buttons on the admin notices in `Generic_AdminNotes.php`, `Cdn_AdminNotes.php`, `UsageStatistics_Plugin_Admin.php`, `Minify_Plugin_Admin.php`, `Extension_Wpml_Plugin_Admin.php`, `Extension_Genesis_Plugin_Admin.php`, `Extension_WordPressSeo_Plugin_Admin.php`, `Extension_NewRelic_AdminNotes.php`. After this PR every one of those buttons gets silently dropped (`Util_Admin::redirect( array(), true )` in the handler) and the notice re-renders on next page load.

Conversely, several keys present in this allowlist (`license.terms`, `license.next_check`, `license.expires`, `common.install_version`, …) are only ever written from `Licensing_*` / `SetupGuide_*` via `Dispatcher::config_state*()->set()` directly, not through the `w3tc_default_config_state*` handlers — so they aren't gated by this allowlist either way.

**Recommended fix:** at minimum, add the 21 missing keys (the grep one-liner I used: `grep -rE "['\"]w3tc_default_config_state(_master|_note)?['\"]" --include="*.php" -A 2 | grep -E "['\"]key['\"]\s*=>" | sed -E "s/.*'key'\s*=>\s*'([^']+)'.*/\1/g" | sort -u`). Better: drop the hardcoded list and use `apply_filters( 'w3tc_state_key_allowlist', [ ... ] )` so extensions register their own dismissable-notice keys — otherwise every new admin notice in core or in any extension will need to touch this file, and out-of-tree extensions can't add notices at all. The threat model here is arbitrary mass-assignment into `w3tc_state`; a namespace-based match like `/\.(hide|show)_note(?:[\._]|$)/` would cover the actual attack surface (dismissable booleans) without the maintenance burden.

While you're here, please add a round-trip test exercising one of the keys above through `?w3tc_action=default_config_state&key=...` and asserting the state value actually persists. `test_is_known_state_key()` only tests the accessor, not the handler's gating decision.

ConfigKeysSchema.php
```diff
@@ -0,0 +1,300 @@
...
+	 * @since X.X.X
+	 *
+	 * @param string|array $key Single-string or [parent, child] compound key.
+	 *
+	 * @return bool
+	 */
+	public static function is_known( $key ) {
+		if ( \is_array( $key ) ) {
+			return true;
+		}
```
cssjoe, line 134: Minor: `is_known()` returns `true` for ANY array, of any shape. The intent (per the doc block) is "compound (`[extension_id, sub_key]`) keys are extension-owned" but a 5-element nested array also passes. In `read_request()` the only producer is `Util_Ui::config_key_from_http_name()` which always yields a 2-element array of strings, so this is fine today — but the schema accessor reads as "trust any array" which is a footgun for future callers.

Suggested tightening: `is_array($key) && count($key) === 2 && is_string($key[0]) && is_string($key[1])`.

Generic_AdminActions_Default.php
```diff
@@ -787,6 +813,14 @@ public function read_request( $config ) {
 			*/
 			$descriptor = apply_filters( 'w3tc_config_key_descriptor', $descriptor, $key );
 
+			// Strict allowlist gate: a key reaches `$config->set()` only if
+			// either (a) it's a compound extension key (the extension owns its
+			// own gate via the filter above), or (b) the schema knows it or
+			// the filter supplied a descriptor for it.
+			if ( ! is_array( $key ) && null === $descriptor && ! ConfigKeysSchema::is_known( $key ) ) {
```
cssjoe, line 820: Minor: the third clause `! ConfigKeysSchema::is_known( $key )` is redundant. For a non-compound key, if `is_known($key)` were true then the `descriptor()` call a few lines above would already have returned non-null, and the filter typically doesn't null out a populated descriptor — so `null === $descriptor` would already be false. The only case the redundancy buys you anything is "filter explicitly returned null for a key that IS in the schema", and in that case the new code lets the write through with no type coercion (matching legacy behavior). Probably intended; just consider simplifying to `null === $descriptor && ! is_array( $key )` plus a code comment, since the current form reads as if there's a third path that doesn't exist.

cssjoe (dismissed):
Reviewed against commit 826111c (PHP 8.3.31). Architectural direction is sound — boundary gating + treating `ConfigKeys.php` as the schema of record + folding non-scalars before casting are the right calls. The follow-up commit (`2d0859f`) addressing Copilot's first round (array→bool coerce, `T_CLOSE_TAG`, real-assignment tracking) was solid.

Two blocking issues found, plus a few smaller items (inline).

## Blocking

### 1. `validate_dbcluster_content()` is bypassable — confirmed RCE on PHP 8.x

The deny-list + "T_STRING followed by `(`" check misses several routes to call-an-expression in modern PHP. Standalone repro (validator function lifted into a probe) accepts all six of these:

```
variable-function-call                     => ACCEPT  $cmd = 'system'; $cmd('id');
namespaced-function-call-php8              => ACCEPT  \system('id');
qualified-namespace-call                   => ACCEPT  NS\bad('id');
arrow-fn-with-var-invoke                   => ACCEPT  $evil = fn($x) => $x('id'); $evil('system');
string-followed-by-paren-direct            => ACCEPT  ("system")("id");
parenthesized-assignment-as-callable       => ACCEPT  ($x = 'system')('id');
```

End-to-end (validate → `file_put_contents` → `require`) executes `system('id')` and prints `uid=...` — re-opens the rt9-219 / rt9-222 RCE primitive this PR claims to close.

Root cause:

- `T_VARIABLE` followed by `(` is not checked (PHP supports variable functions).
- On PHP 8.0+ `\system` / `NS\bad` tokenize as a single `T_NAME_FULLY_QUALIFIED` / `T_NAME_QUALIFIED`, not `T_STRING`, so the bareword check never fires. W3TC supports 7.2.5–8.3, so PHP 8 is in scope.
- `T_FN` (PHP 7.4+) is not in `$forbidden_tokens`.
- `)` and `T_CONSTANT_ENCAPSED_STRING` followed by `(` aren't covered at all.

Suggested fix: invert to an allowlist for `(` — walk back past `T_WHITESPACE`/`T_COMMENT`/`T_DOC_COMMENT` and reject unless the previous significant token is exactly `T_STRING` with value `array`. That single rule retires every call shape without enumerating token types per PHP version. Also add `T_FN` to the deny list — the dbcluster file should never declare callables. Tests in `class-w3tc-dbcluster-validator-test.php` need the six payloads above as new entries in `test_validate_rejects_code_execution_primitives()`.

### 2. `$state_key_allowlist` is missing every dismissable-notice key actually used in the codebase

A grep across the plugin for every `'key' => '...'` paired with `'w3tc_default_config_state{,_master,_note}' => 'y'` (the `Util_Ui::button_hide_note2()` / `Util_Ui::url()` shapes that drive the three handlers) yields **21 keys, zero overlap** with the new allowlist:

```
cdn.hide_note_no_curl                        common.show_note.flush_posts_needed
cdn.show_note_cdn_reupload                   common.show_note.flush_statics_needed
cdn.show_note_cdn_upload                     common.show_note.nginx_restart_required
cdn.show_note_theme_changed                  common.show_note.plugins_updated
cdn.show_note_wp_upgraded                    genesis.theme.hide_note_suggest_activation
common.hide_note_debug_enabled               minify.show_note_minify_error
common.hide_note_no_permalink_rules          newrelic.hide_note_pageload_slow
common.hide_note_no_zlib                     wordpress_seo.hide_note_suggest_activation
common.hide_note_stats_enabled               wpml.hide_note_language_negotiation_type
common.hide_note_wp_content_permissions      wpml.hide_note_suggest_activation
common.hide_note_zlib_output_compression
```

These are the keys behind the "X" / "Hide this notice" buttons in `Generic_AdminNotes.php`, `Cdn_AdminNotes.php`, `Extension_*_Plugin_Admin.php`, etc. After this PR, every one of those buttons hits the handler, gets silently dropped, and the notice re-renders on next load.

Conversely the allowlist contains `license.terms`, `license.next_check`, `license.expires`, `common.install_version`, etc. — those are written from `Licensing_*` / `SetupGuide_*` directly via `Dispatcher::config_state*()->set()`, not through these handlers, so they're not protected by this gate either way.

Suggested fix: at minimum, audit + add the 21 missing keys. Better: replace the hardcoded list with `apply_filters( 'w3tc_state_key_allowlist', [...] )` so extensions can register their own dismissable-notice keys — otherwise every new admin notice in core or in any extension will need to touch `ConfigKeysSchema.php`, and out-of-tree extensions can't add notices at all. A namespace pattern like `/\.(hide|show)_note(?:[\._]|$)/` would also cover the actual attack surface (dismissable booleans).

While you're there, please add a round-trip test: simulate `?w3tc_action=default_config_state&key=...&value=...` for one of the keys used in `Generic_AdminNotes.php` and assert `Dispatcher::config_state()->get($key)` reflects the new value. `test_is_known_state_key()` only exercises the schema accessor, not the handler.

## Non-blocking — see inline comments

- `ConfigKeysSchema::is_known()` returns `true` for ANY array (compound-key handling is too permissive at the schema level).
- Dead clause in the `read_request()` gate.
- `<?php` open-tag check is case-sensitive (PHP accepts mixed-case open tags).
- `Config::import()` runs `ConfigCompiler::upgrade()` before the gate — worth a docblock note.
- 0% patch coverage on the actual gates (`Config.php`, `Generic_AdminActions_Default.php`) per Codecov; the new tests cover the accessor and the validator in isolation but not the boundaries that call them.

## Repro artifacts

The two standalone files I used to confirm the dbcluster bypass (lift the validator function into a probe, then `validate → file_put_contents → require` to run `system('id')`) are available on request if you'd like them vendored as test fixtures.

cssjoe:
Closing this PR. Continued work on this fix is being handled through a private review process. A new public PR will be opened when the change is ready to merge. Thanks to everyone who reviewed here.

PR #1321, opened 2026-05-13
Encrypt at-rest credentials and harden W3TC cookies

jacobd91:
https://imh-internal.atlassian.net/jira/software/c/projects/ENG7/boards/863/backlog?selectedIssue=ENG7-3011

  Closes the RT9 `crypto-secrets-cookies` group: encrypt at-rest CDN
  credentials in `master.php`, replace `md5(NONCE_KEY . $role)` with an
  HMAC-SHA256 cookie name (with one-release back-compat read), set
  secure/httponly/samesite=Lax on every `w3tc_*` cookie, render
  credential form inputs as placeholders (no plaintext in HTML
  `value=`), re-enable TLS cert verification on the License API.

  **Closes 5 findings (3 High, 2 Medium)** and **contributes to
  CHAIN-013** (sslverify on outbound calls — license half).

  | Bead    | Severity | Finding                 |
|---------|----------|--------------------------------------------------------------------------------------------------------------------|
  | rt9-14  | High     | A02:2021 — CDN API credentials stored plaintext in W3TC config                |
  | rt9-13  | High     | A02:2021 — TLS certificate verification disabled on outbound license calls (license half;image-service/ttfb in SSRF PR) |
  | rt9-11  | High     | A07:2021 — MD5 cookie name from `md5(NONCE_KEY . $role)` (broken hash for cache-key auth marker)                 |
  | rt9-19  | Medium   | A02:2021 — Configuration secrets reflected into HTML attribute values on every settings-pagerender                |
  | rt9-10  | Medium   | A07:2021 — Cookies set without Secure / HttpOnly / SameSite flags                   |

  ## What changed

  | File                                       | Net     | Role                  |
|--------------------------------------------|--------:|-----------------------------------------------------------------------------------|
  | `Util_Crypto.php` *(new)*                  | +198    | AES-256-CBC + HMAC envelope, `enc:v1:...` format, legacyplaintext passthrough    |
  | `Util_Cookie.php` *(new)*                  | +154    | Array-form `setcookie` wrapper + HMAC role-cookie name + legacyreader            |
  | `ConfigKeys.php`                           | +19     | 19 credential-typed keys flagged `'flags' => array('secret' =>true)`             |
  | `Config.php`                               | +112    | `encrypt_secrets()` / `decrypt_secrets()` + `secret_keys()`cache                 |
  | `ConfigCompiler.php`                       | +13     | `encrypt_secrets()` before write; decrypt master before childseal-compare        |
  | `Generic_AdminActions_Default.php`         | +16     | Empty POST for a secret-flagged key preserves the existing value                  |
  | `Util_Ui.php`                              | +138    | `secret_input()` helper + `is_secret_config_key()` +`control2()` auto-mask       |
  | `Licensing_Core.php`                       | −4      | Removed all four `'sslverify' => false`                  |
  | Cookie writers (4 files)                   | ~30     | `Util_Environment::set_preview`,`PgCache_Plugin::on_logout/on_login`, `Mobile_Referrer`, `Generic_Plugin` role-cookies |
  | Cookie name reads (2 files)                | ~25     | `PgCache_ContentGrabber::_can_read_cache` +`PgCache_Environment` reject-cookie list both walk new + legacy |
  | Credential views (`inc/options/cdn/*.php` + `general.php`) | ~120 | Route every secret input through`Util_Ui::secret_input()` |
  | `tests/admin/class-w3tc-util-crypto-test.php` *(new)` | +175 | 9 tests covering round-trip, tamper, legacy passthrough, no-double-wrap |
  | `tests/admin/class-w3tc-util-cookie-test.php` *(new)` | +90  | 5 tests covering role-cookie name shape + determinism +new-vs-legacy |

  ## Architectural decisions worth your eye on review

  ### 1. Two new helper classes as chokepoints

  Put crypto in `Util_Crypto.php` and cookie writes in `Util_Cookie.php`. Every cookie write and every credential save MUST
   route through these — otherwise a future contributor adds a `setcookie( 'w3tc_X', ... )` without flags, or a new
  credential key without the encrypt path, and the protection silently regresses. Class-as-chokepoint is more maintainable
  than "remember to pass these args every time".

  ### 2. Encrypt-on-write, decrypt-on-load (no audit of read sites)

  The in-memory `$this->_data` stays in plaintext; only the disk representation is enveloped. Every existing
  `$config->get_string('cdn.s3.secret')` call site continues to receive plaintext — no audit of the ~700 read sites needed.

  master.php on disk:    cdn.s3.secret = 'enc:v1:6URx3BEi...'
  Config::load():        decrypt_secrets($data); → cdn.s3.secret = 'wJalrXUtn...'
  $config->get_string(): cdn.s3.secret → 'wJalrXUtn...' (plaintext, unchanged from pre-fix)
  Config::save():        plaintext flows into ConfigCompiler::save()
  ConfigCompiler:        encrypt_secrets($data); → master.php gets 'enc:v1:...'

  Tamper failures (HMAC mismatch, malformed base64) collapse to empty string, so the admin form shows "credential needs
  re-entry" rather than the literal `false`.

  ### 3. One-release cookie-name back-compat window

  `Generic_Plugin` writes the new HMAC name only. `PgCache_ContentGrabber` and the `PgCache_Environment` reject-cookie list
   (both call sites) accept BOTH names. Browsers carrying a pre-upgrade `w3tc_logged_<MD5>=1` cookie continue to bypass
  cache during the transition.

  Drop the legacy reader and the matching reject-cookie entry in the release AFTER this one. Comments inline mark the drop
  sites.

  ### 4. Empty-POST-preserves-secret rule (the load-bearing companion)

  Without this, the `value=""` placeholder pattern wipes every CDN credential on the first Settings save. Easiest footgun
  in the whole PR to miss in review:

  ```
  php
  // Generic_AdminActions_Default::read_request()
  if (
      is_array( $descriptor )
      && ! empty( $descriptor['flags']['secret'] )
      && ( ! is_string( $request_value ) || '' === $request_value )
  ) {
      continue;  // empty POST for a secret means "no change", not "wipe it".
  }
  ```

  ### 5. Salt rotation invalidates secrets — by design

  If an operator rotates secure_auth (or auth) salts in wp-config.php, every previously-encrypted secret will fail to
  decrypt. That matches WordPress's own behaviour for cookie-signed sessions — the salt is the master secret. Operators
  rotating salts must re-enter credentials. The release notes should call this out alongside the existing salt-rotation
  guidance.

  ### 6. AES-256-CBC + HMAC, not AES-256-GCM

  GCM would be preferred for new code, but openssl_encrypt's GCM mode requires PHP 7.1+ AND an OpenSSL build that exposes
  it — uneven in the PHP-7.2.5 → 8.3 support window. CBC + HMAC-SHA256 (encrypt-then-MAC, with a hash_equals check before
  openssl_decrypt) is universally supported and gives the same authenticated-encryption guarantee at this threat model.
  derive_key() uses HMAC-over-constant-string-keyed-by-salt so a future salt re-use can't accidentally produce the same
  envelope key.

  ### 7. Util_Ui::control2() auto-mask via flags.secret

  Util_Ui::secret_input() is the explicit helper for direct calls. But many credential inputs are rendered through
  Util_Ui::config_item() → control2(). So control2() checks is_secret_config_key() on textbox-typed items and routes
  through secret_input() automatically — future credential fields added to the schema are masked by being flagged, no
  template edit needed.

  Test plan

  Automated (passing)

  1. PHP syntax check on every changed file
  php -l Util_Crypto.php Util_Cookie.php Config.php ConfigCompiler.php \
         ConfigKeys.php Util_Ui.php Util_Environment.php Generic_Plugin.php \
         Generic_AdminActions_Default.php PgCache_Plugin.php \
         PgCache_ContentGrabber.php PgCache_Environment.php \
         Mobile_Referrer.php Licensing_Core.php \
         tests/admin/class-w3tc-util-crypto-test.php \
         tests/admin/class-w3tc-util-cookie-test.php

  2. Lint everything
  yarn run php-lint

  3. Full PHPUnit suite
  ./vendor/bin/phpunit
  Expected: 92 tests, 331 assertions, all green

  4. New tests in isolation
  ./vendor/bin/phpunit --filter 'W3tc_Util_Crypto_Test|W3tc_Util_Cookie_Test'
  Expected: 13 tests, 45 assertions

  5. WordPress coding-standards check
  ./vendor/bin/phpcs Util_Crypto.php Util_Cookie.php Config.php ConfigCompiler.php

  Manual on staging (for the reviewer)

  A. At-rest encryption — round-trip

  1. Fresh W3TC install. Performance → CDN → enable an engine and enter a credential (e.g. S3 access key + secret).
  2. Save.
  3. Read wp-content/w3tc-config/master.php. Confirm:
    - cdn.s3.key and cdn.s3.secret appear as 'enc:v1:...' strings.
    - The base64 body of each is different (fresh IV per encrypt).
  4. Hit any frontend page; confirm CDN URLs render correctly (the read path decrypts in-memory).
  5. Performance → Dashboard → Empty all caches; confirm no notices in debug.log.

  B. At-rest encryption — placeholder round-trip

  1. With the credentials saved from step A, reload the CDN settings page. Confirm:
    - View source → every credential input is value="".
    - The visible field shows •••••••• as placeholder (proof that "something is configured" without leaking the value).
  2. Save the Settings page without re-entering anything. Read master.php again. Confirm cdn.s3.secret is unchanged (the
  empty-POST-preserves-secret rule fired).
  3. Now enter a new value into cdn.s3.secret and save. Confirm master.php has a new enc:v1:... blob.

  C. At-rest encryption — tamper detect

  Read the encrypted value from master.php, flip a byte in the middle, write back.
  ENC=$(grep -oE "'enc:v1:[^']+'" wp-content/w3tc-config/master.php | head -1)
  ... manually edit one base64 char ...
  Reload the CDN settings page.
  Expected: the cdn.s3.secret field renders with an empty placeholder (no ••••••••), because the tamper-detect returned
  false and the value collapsed to ''. The admin can re-enter and save.
  
  D. Cookie flags

  1. Browse the site over HTTPS (logged in as admin).
  2. Open browser DevTools → Network → click any request → Response Headers.
  3. Confirm every Set-Cookie: w3tc_* line includes Secure; HttpOnly; SameSite=Lax.
  4. Repeat over plain HTTP — Secure is absent (correct: would break http-only test installs).

  E. Cookie name migration + back-compat

  1. Before upgrade: log in as admin. DevTools → Application → Cookies. Find w3tc_logged_<MD5> (32 hex chars). Note the
  value.
  2. Upgrade to this branch.
  3. Log out + log back in. Confirm:
    - A new cookie w3tc_logged_<HMAC> appears (also 32 hex chars; DIFFERENT from the MD5 one).
  4. Inject the legacy MD5 cookie back into the browser manually (DevTools → set a cookie with the original MD5 name and 1
  value). Hit the front page.
    - Confirm the page is served fresh (no cache hit) — the back-compat reader in _can_read_cache recognised the legacy
  cookie and bypassed cache.
  5. Repeat with only the new HMAC cookie present — should also bypass cache.

  F. License API TLS verify

  1. Performance → General Settings → Licensing → enter a valid license key and click Activate.
  2. Confirm the activation succeeds (server is reachable over HTTPS with a valid cert).
  3. As an MITM test, point the licensing API URL at a host with a self-signed cert (e.g. via /etc/hosts). Confirm the
  activation fails (previously would have succeeded with sslverify => false).

  G. nginx / Apache reject-cookie config

  If you have pgcache.engine = file_generic and pgcache.reject.logged_roles = true:

  1. Regenerate the page-cache config (Performance → General Settings → Save All Settings — re-emits .htaccess /
  nginx.conf).
  2. Inspect .htaccess (or nginx.conf). Confirm the reject-cookie list contains BOTH w3tc_logged_<HMAC> AND
  w3tc_logged_<MD5> entries for each rejected role (back-compat window).
  3. After the next release drops the legacy form, only the HMAC entries should remain.

  Verification

  Static code review only — the RT9 engagement lab is decommissioned and per-finding proof.sh scripts depend on lab
  credentials excluded from the audit handoff. Each fix was confirmed by reading the finding's Vulnerable Code +
  Remediation sections in tmp/sec-ideas/crypto-secrets-cookies.md, inspecting the post-fix file, and running the new
  WP_UnitTestCase coverage end-to-end (92/92 pass).

  The crypto envelope was smoke-tested directly against:
  - 8 round-trip cases (short ASCII, AWS-key shape, user:pass, spaces, quotes, 4KB string, UTF-8, single char)
  - Tamper detection (flipped byte → false; truncation → false; malformed base64 → false)
  - Legacy plaintext passthrough
  - No double-wrap on already-enveloped input
  - Fresh IV per call (same plaintext → different envelope)

  Staging note (important)

  Util_Crypto.php and Util_Cookie.php are new files. git add -u and git commit -a skip new files — stage them explicitly:

  git add Util_Crypto.php Util_Cookie.php Config.php ConfigCompiler.php \
          ConfigKeys.php Util_Ui.php Util_Environment.php Generic_Plugin.php \
          Generic_AdminActions_Default.php PgCache_Plugin.php \
          PgCache_ContentGrabber.php PgCache_Environment.php \
          Mobile_Referrer.php Licensing_Core.php \
          inc/options/cdn/ inc/options/general.php \
          tests/admin/class-w3tc-util-crypto-test.php \
          tests/admin/class-w3tc-util-cookie-test.php

  If the new helpers aren't in the commit, every page request will fatal with Class W3TC\Util_Crypto not found
  (Config::load → decrypt_secrets) or Class W3TC\Util_Cookie not found (any login/logout). Same gotcha pattern as the auth
  PR's Util_Nonce.php, the command-injection PR's Util_Java.php, and the file-inclusion PR's Util_Extension.php.

  Salt rotation note (operator-facing — should land in CHANGELOG)

  If an operator rotates secure_auth (or auth) salts in wp-config.php after this PR is deployed:

  - Every encrypted credential in master.php will fail to decrypt. The admin form will show empty credential fields;
  CDN/FTP/license operations will fail until the credentials are re-entered.
  - Every active cache-bypass cookie (w3tc_logged_<HMAC>) will be invalidated. Logged-in admins will see the cached version
   of pages until they log in again.

  This matches WordPress's own salt-rotation behaviour (cookie-signed sessions invalidate). It's documented inline in
  Util_Crypto.php and should be added to the release notes.

  Out of scope (other audit groups, separate PRs)

  - TLS verify for image-service / TTFB outbound (Extension_ImageService_Api.php, Util_Http::ttfb) — sec-fix-ssrf PR
  (overlap with rt9-13; coordinated).
  - Google Drive OAuth state nonce + return-handler trust issues (rt9-150 / rt9-151 / rt9-215) — separate OAuth-hardening
  pass.
  - Memcached/Redis password POST-only restriction (rt9-12) — sec-fix-missing-auth PR.
  - Swarmify / NewRelic mass-assignment hardening (rt9-216 / rt9-217) — sec-fix-mass-assignment PR.
  - Google API library replacement (rt9-43) — dependency-upgrade pass.
  - Extension-level credentials (cloudflare.key, bunnycdn.api_key, etc.) — not in the static ConfigKeys.php schema;
  separate per-extension pass.

### DIFF
--- a/Config.php
+++ b/Config.php
@@ -623,6 +623,8 @@ class Config {
 			}
 		}
 
+		self::decrypt_secrets( $data );
+
 		$this->_data     = $data;
 		$this->_compiled = false;
 	}
@@ -635,7 +637,123 @@ class Config {
 	private function load_full() {
 		$c = new ConfigCompiler( $this->_blog_id, $this->_preview );
 		$c->load();
-		$this->_data     = $c->get_data();
+		$data = $c->get_data();
+		self::decrypt_secrets( $data );
+		$this->_data     = $data;
 		$this->_compiled = true;
+	}
+
+	/**
+	 * Returns the set of config keys declared as `secret` in
+	 * ConfigKeys.php — credential-typed keys whose stored value is
+	 * encrypted with {@see Util_Crypto::envelope_encrypt()}.
+	 *
+	 * The schema include is cached for the request, so this is cheap
+	 * to call from both the load and save paths.
+	 *
+	 * @since X.X.X
+	 *
+	 * @return array<string,true> Map of secret key name → true.
+	 */
+	public static function secret_keys() {
+		static $cache = null;
+
+		if ( null !== $cache ) {
+			return $cache;
+		}
+
+		$keys = array();
+		include W3TC_DIR . '/ConfigKeys.php';
+
+		$cache = array();
+		if ( is_array( $keys ) ) {
+			foreach ( $keys as $name => $descriptor ) {
+				if (
+					is_array( $descriptor )
+					&& isset( $descriptor['flags'] )
+					&& is_array( $descriptor['flags'] )
+					&& ! empty( $descriptor['flags']['secret'] )
+				) {
+					$cache[ $name ] = true;
+				}
+			}
+		}
+
+		return $cache;
+	}
+
+	/**
+	 * Decrypts every secret-flagged key in-place inside `$data`.
+	 *
+	 * Legacy plaintext values pass through unchanged (the envelope
+	 * helper is a no-op on non-`enc:v1:` strings), so an existing
+	 * install upgrades transparently — values stay readable until the
+	 * next `save()` re-wraps them.
+	 *
+	 * Tampered envelopes (bad HMAC, malformed base64) decrypt to
+	 * `false`; we collapse that to an empty string so the downstream
+	 * config consumer sees "credential needs re-entry" rather than the
+	 * literal `false` (which `get_string()` would coerce to `""` anyway
+	 * but via a non-obvious code path).
+	 *
+	 * @since X.X.X
+	 *
+	 * @param array $data Configuration data (modified in place).
+	 *
+	 * @return void
+	 */
+	public static function decrypt_secrets( &$data ) {
+		if ( ! is_array( $data ) || ! class_exists( '\W3TC\Util_Crypto' ) ) {
+			return;
+		}
+
+		$secret_keys = self::secret_keys();
+		if ( empty( $secret_keys ) ) {
+			return;
+		}
+
+		foreach ( $secret_keys as $key => $_ ) {
+			if ( ! isset( $data[ $key ] ) || ! is_string( $data[ $key ] ) ) {
+				continue;
+			}
+			$plain = Util_Crypto::envelope_decrypt( $data[ $key ] );
+			if ( false === $plain ) {
+				$data[ $key ] = '';
+			} else {
+				$data[ $key ] = $plain;
+			}
+		}
+	}
+
+	/**
+	 * Encrypts every secret-flagged key in-place inside `$data`.
+	 *
+	 * Called by {@see ConfigCompiler::save()} before writing
+	 * `master.php`. Non-secret keys are untouched. Already-enveloped
+	 * values are not double-wrapped (the envelope helper detects its
+	 * own prefix and short-circuits).
+	 *
+	 * @since X.X.X
+	 *
+	 * @param array $data Configuration data (modified in place).
+	 *
+	 * @return void
+	 */
+	public static function encrypt_secrets( &$data ) {
+		if ( ! is_array( $data ) || ! class_exists( '\W3TC\Util_Crypto' ) ) {
+			return;
+		}
+
+		$secret_keys = self::secret_keys();
+		if ( empty( $secret_keys ) ) {
+			return;
+		}
+
+		foreach ( $secret_keys as $key => $_ ) {
+			if ( ! isset( $data[ $key ] ) || ! is_string( $data[ $key ] ) || '' === $data[ $key ] ) {
+				continue;
+			}
+			$data[ $key ] = Util_Crypto::envelope_encrypt( $data[ $key ] );
+		}
 	}
 }
--- a/ConfigCompiler.php
+++ b/ConfigCompiler.php
@@ -265,12 +265,25 @@ class ConfigCompiler {
 			$master = new ConfigCompiler( 0, $this->_preview );
 			$master->load();
 
+			// The master config is read from disk as `enc:v1:...` for
+			// secret keys; the child config we're saving holds those
+			// keys as plaintext. Decrypt the master copy so the
+			// seal-comparison below is a like-for-like check, otherwise
+			// every child save would think every secret had changed.
+			Config::decrypt_secrets( $master->_data );
+
 			foreach ( $this->_data as $key => $value ) {
 				if ( ! self::child_key_sealed( $key, $master->_data, $this->_data ) ) {
 					$data[ $key ] = $this->_data[ $key ];
 				}
 			}
 		}
+
+		// Encrypt credential-typed keys (`flags.secret => true`) before
+		// they hit master.php. The in-memory `$this->_data` stays in
+		// plaintext so callers reading their values during the same
+		// request continue to see decrypted strings.
+		Config::encrypt_secrets( $data );
 
 		ConfigUtil::save_item( $this->_blog_id, $this->_preview, $data );
 	}
--- a/ConfigKeys.php
+++ b/ConfigKeys.php
@@ -31,10 +31,12 @@ $keys = array(
 	'cluster.messagebus.sns.api_key'                       => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ), // AWS SNS API key — encrypted at rest.
 	),
 	'cluster.messagebus.sns.api_secret'                    => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cluster.messagebus.sns.topic_arn'                     => array(
 		'type'    => 'string',
@@ -1277,6 +1279,7 @@ $keys = array(
 	'cdn.ftp.pass'                                         => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ), // FTP password — encrypted at rest.
 	),
 	'cdn.ftp.path'                                         => array(
 		'type'    => 'string',
@@ -1305,6 +1308,7 @@ $keys = array(
 	'cdn.ftp.privkey'                                      => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ), // SFTP private key — encrypted at rest.
 	),
 	'cdn.google_drive.client_id'                           => array(
 		'type'    => 'string',
@@ -1313,6 +1317,7 @@ $keys = array(
 	'cdn.google_drive.refresh_token'                       => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cdn.google_drive.folder.id'                           => array(
 		'type'    => 'string',
@@ -1329,10 +1334,12 @@ $keys = array(
 	'cdn.s3.key'                                           => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ), // S3 access key ID — encrypted at rest.
 	),
 	'cdn.s3.secret'                                        => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ), // S3 secret key.
 	),
 	'cdn.s3.bucket'                                        => array(
 		'type'    => 'string',
@@ -1361,10 +1368,12 @@ $keys = array(
 	'cdn.cf.key'                                           => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cdn.cf.secret'                                        => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cdn.cf.bucket'                                        => array(
 		'type'    => 'string',
@@ -1393,10 +1402,12 @@ $keys = array(
 	'cdn.cf2.key'                                          => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cdn.cf2.secret'                                       => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cdn.cf2.id'                                           => array(
 		'type'    => 'string',
@@ -1417,6 +1428,7 @@ $keys = array(
 	'cdn.rscf.key'                                         => array(
 		'type'    => 'string',
 		'default' => '',
+		'flags'   => array( 'secret' => true ),
 	),
 	'cdn.rscf.location'                                    => array(
 		'type'    => 'string',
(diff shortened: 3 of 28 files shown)


cssjoe:
## Review

Strong, well-architected security PR. The threat model is correct and several implementation choices are noticeably above-average — encrypt-then-MAC with independent keys, the empty-POST-preserves-secret rule, the one-release legacy-cookie window, and the PHP 7.2 `SameSite` smuggling. All five Copilot review comments are addressed by `942527b`. Approve from my side; the items below are nice-to-haves and notes, not blockers.

### Documented behaviour to flag in release notes

**`Config::export()` writes plaintext secrets to JSON.** `$this->_data` is decrypted-in-memory, so `wp_json_encode( $this->_data )` serializes every `flags.secret` key as plaintext into the export. Same for `Cli.php::export()`. Same threat surface as a `master.php` disclosure, just via a different channel.

Out of scope for this PR per discussion — we accept exports as plaintext for now — but worth one line in the release notes alongside the existing salt-rotation guidance so operators know the export endpoint is a parallel exfil primitive.

### Nice-to-haves (non-blocking)

1. **PHPCS warns 3× on `@\openssl_*` in `Util_Crypto.php`** (lines 87, 156, 211 — `WordPress.PHP.NoSilencedErrors.Discouraged`). The `@` is correct — return values are explicitly tested for `false`, and on `openssl_decrypt` it's load-bearing to suppress the `OPENSSL_R_BAD_DECRYPT` warning that would otherwise flood `debug.log` on every tamper-detect input. Suggest a one-line `// phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged -- ...` above each call rather than leaving warnings in the lint output.

2. **Cookie security-flag flip is a behaviour change worth a release-note line.** `w3tc_preview`, `w3tc_referrer`, `w3tc_logged_*`, and `w3tc_logged_out` previously had no flags and now carry `HttpOnly` + `SameSite=Lax` + `Secure` on HTTPS. I verified the plugin's own JS doesn't read any of them from `document.cookie` (no internal regression), but third-party integrations reading `w3tc_referrer` client-side will stop seeing it. One line in the changelog covers it.

3. **`Util_Ui::secret_input()` docblock could be clearer that `type=text` still gets `value=""` + dot-placeholder treatment.** The current "`password` (default) or `text`" wording reads like `text` re-enables `value=`. It doesn't — `text` just renders the input as text-type so operators can read back e.g. an Access Key ID without unmasking. SNS API key, S3 access key ID, CloudFront access key ID, and the EDD license key all rely on this.

4. **Inconsistent inline annotation in `ConfigKeys.php`.** Five of the nineteen new `'flags' => array( 'secret' => true )` lines get a trailing comment (e.g. `// FTP password — encrypted at rest.`), the rest don't. Cosmetic — pick all-or-none.

5. **Brief code-comment for `Config::set_defaults()`.** It instantiates `ConfigCompiler` without calling `load()`, so no decrypt is needed there (constructor-only defaults). A one-liner cross-reference saves the next reader a "wait, why no `decrypt_secrets()` here?" detour.

### Things done well — please keep these load-bearing as-is

- **Encrypt-on-write / decrypt-on-load** with `$this->_data` always plaintext. Avoids needing to audit ~700 `get_string()` call sites. Worth a `@security-invariant` comment on `Config::decrypt_secrets()` so a future contributor doesn't try to "fix" the in-memory representation to match disk.
- **Encrypt-then-MAC with independent keys** (`derive_key('enc')` vs `derive_key('mac')`). Critical to keep — a future "simplify" refactor collapsing them back to one key would reintroduce the design smell. `test_derive_key_uses_distinct_purposes` correctly pins both the determinism and not-equal properties via reflection.
- **`Generic_AdminActions_Default::read_request()` empty-POST-preserves-secret rule.** The PR description calls this "the load-bearing companion" and it really is — easy to delete in a future refactor because it looks redundant. A code-comment cross-reference back to `Util_Ui::secret_input()` would harden it against accidental removal.
- **PHP 7.2 `SameSite` smuggling via path.** `composer.json`'s platform pin is `7.2.5` and the 7.3+ options-array form silently drops the cookie on 7.2. The `PHP_VERSION_ID >= 70300` guard with `$path . '; samesite=Lax'` fallback is the right shape for the supported PHP window. Don't let a future "bump min PHP to 7.4" effort delete the fallback without first updating `composer.json`.
- **One-release legacy-cookie reader AND reject-cookie list.** Both `PgCache_ContentGrabber::_can_read_cache()` and `PgCache_Environment` (the reject-cookie list emitted into `.htaccess`/`nginx.conf`) walk both names. Easy to miss the second site on a casual read — the inline `// Drop in the next release` comments correctly mark both drop sites.
- **`Mobile_Referrer` set/clear path consistency** (the Copilot follow-up). The browser keys cookies on `(name, path, domain)`, so set-with-`/` then clear-with-`COOKIEPATH` would leak a permanent zombie cookie on sub-directory installs. Subtle and important.

### Sanity-checks I ran

- `php -l` on all 11 modified PHP files → clean.
- `phpcs --standard=phpcs.xml` on the 11 changed runtime files → 3 warnings in `Util_Crypto.php` (the silenced-error nit above) + 1 pre-existing alignment warning in `Generic_Plugin.php` (blamed to a 2025-11-25 commit, unrelated).
- `grep` confirmed `ConfigUtil::save_item()` is only called from `ConfigCompiler::save()` — encrypt path is universal.
- `grep` confirmed no plugin JS reads `w3tc_preview`/`w3tc_referrer`/`w3tc_logged_*` from `document.cookie` — `HttpOnly` is safe for the plugin's own bundle.
- Verified `Config::secret_keys()`'s request-scoped `static $cache = null` correctly caches the schema include across `encrypt_secrets()`/`decrypt_secrets()`/`is_secret_config_key()` calls.

LGTM with the release-notes additions.

cssjoe (approved):

PR #1333, opened 2026-05-18
Add w3tc_cache_allowed_classes filter and treat unknown objects as cache miss

jacobd91:
https://imh-internal.atlassian.net/browse/ENG7-3003


cssjoe (changes requested):
See Copilot comments.

cssjoe (approved):

PR #1353, opened 2026-07-01
Fix for htaccess -multiviews error

jacobd91:
2.10.0 introduced a total-site-outage regression on Apache/LiteSpeed. The page-cache CORE rules block — written to the site-root .htaccess (Util_Rule::get_pgcache_rules_core_path()), which Apache parses on every request — began emitting an unconditional Options -MultiViews:

# BEGIN W3TC Page Cache core
Options -MultiViews          ← added in 2.10.0
<IfModule mod_rewrite.c>
...

An Options directive in .htaccess requires the vhost to grant AllowOverride Options (or All). Many secure/shared hosts use AllowOverride FileInfo — which permits W3TC's RewriteRules (so rewrites kept working) but forbids Options. On those hosts Apache aborts .htaccess parsing with Option MultiViews not allowed here and returns HTTP 500 for every URL, taking down both the front end and /wp-admin/. Affected: Page Cache enabled with method Disk: Enhanced (file_generic).

Reported on the support forum by multiple users (all Apache 2.4, error .htaccess:Option MultiViews not allowed here); recovery required manual .htaccess/plugin edits over SSH/FTP.

Root cause / rationale

The root-level directive was added in the 2.10.0 cycle as a QA/functional-consistency change (commit "Fix tests"), alongside a new disk-enhanced-probe.php flag that asserted its presence — not a security fix. It carries no RT9-NNN reference, maps to no audit finding, and is not load-bearing for any security RewriteCond. W3TC shipped for years (through 2.9.x) without this directive in the root file. QA missed the regression because every sandbox vhost uses AllowOverride All, which makes the directive legal.

▎ Note: an Options directive in .htaccess is fundamentally unenforceable on the exact hosts that crash (their AllowOverride forbids it), and an <IfModule> wrapper cannot suppress the error (it originates from AllowOverride at parse time, not a missing module). So there is no safe "keep it" variant for those servers.

Changes

- PgCache_Environment.php — remove Options -MultiViews from the CORE block. The block now goes directly from the marker to <IfModule mod_rewrite.c>, restoring pre-2.10.0 behavior.
- qa/plugins/disk-enhanced-probe.php — drop the now-obsolete htaccess_multiviews probe flag (nothing in the JS suite asserted on it).

The guarded, subdirectory-scoped Options -MultiViews directives (PgCache_Environment.php cache-dir block and Minify_Environment.php, both gated by $compatibility || $disk_enhanced_apache) are intentionally left in place — they write to subdir .htaccess files and were not the outage cause.

Testing instructions

Repro (before this fix / on 2.10.0):
1. Apache 2.4 vhost with a restrictive override, e.g. AllowOverride FileInfo (grants mod_rewrite, not Options).
2. W3TC → Page Cache enabled, method Disk: Enhanced; Save all settings & purge caches.
3. Load any front-end URL or /wp-admin/ → HTTP 500; error log shows .htaccess: Option MultiViews not allowed here.

Verify the fix (this branch):
1. Deploy the branch to the same environment.
2. W3TC → Save all settings & purge caches (regenerates the root .htaccess).
3. Inspect the site-root .htaccess: the # BEGIN W3TC Page Cache core block no longer contains Options -MultiViews (marker → <IfModule mod_rewrite.c>).
4. Front end and /wp-admin/ return HTTP 200; no Option MultiViews not allowed here in the error log.
5. Confirm Disk: Enhanced caching still works — request a cached page twice and verify it's served from cache (cache file present under wp-content/cache/page_enhanced/..., and served HTML/X-W3TC markers as expected).

Regression check (permissive server, AllowOverride All):
- Disk: Enhanced page caching functions exactly as before, minus the Options -MultiViews line.

### DIFF
--- a/PgCache_Environment.php
+++ b/PgCache_Environment.php
@@ -744,7 +744,6 @@ class PgCache_Environment {
 
 		$rules  = '';
 		$rules .= W3TC_MARKER_BEGIN_PGCACHE_CORE . "\n";
-		$rules .= "Options -MultiViews\n";
 		$rules .= "<IfModule mod_rewrite.c>\n";
 		$rules .= "    RewriteEngine On\n";
 		$rules .= '    RewriteBase ' . $rewrite_base . "\n";
--- a/qa/plugins/disk-enhanced-probe.php
+++ b/qa/plugins/disk-enhanced-probe.php
@@ -144,7 +144,6 @@ echo wp_json_encode(
 		'apache_cond_exists'     => ( '' !== $apache_cond_resolved && file_exists( $apache_cond_resolved ) ),
 		'htaccess_core'          => ( false !== strpos( $htaccess, W3TC_MARKER_BEGIN_PGCACHE_CORE ) ),
 		'htaccess_uri'           => ( false !== strpos( $htaccess, 'W3TC_URI_PATH_SLASH' ) ),
-		'htaccess_multiviews'    => ( false !== strpos( $htaccess, 'Options -MultiViews' ) ),
 		'apache_prefix_sample'   => $apache_prefix,
 	)
 );


cssjoe (approved):

cssjoe (approved):
## Review summary

**Verdict: Approve**

Verified the diff, history, and CI. This is the right fix — minimal, restores pre-2.10.0 behavior, and targets the actual outage vector.

### Root cause — confirmed

- `Options -MultiViews` in the **site-root** Page Cache core block (via `Util_Rule::get_pgcache_rules_core_path()`) requires `AllowOverride Options`. On hosts with `AllowOverride FileInfo` (common on shared/secure Apache), Apache aborts `.htaccess` parsing on every request → HTTP 500 for the whole site.
- Pre-2.10.0, the core block went directly from the marker to `<IfModule mod_rewrite.c>`; the directive was introduced in the 2.10.0 cycle. An `<IfModule>` wrapper cannot help here — this is an AllowOverride grant issue at parse time.

### Change scope — correct

- Removing the unconditional root-level directive restores the proven 2.9.x layout without touching the rewrite rules themselves.
- The guarded, subdirectory-scoped `Options -MultiViews` in the cache-dir block (`rules_cache_generate_apache`, gated by `$compatibility || $disk_enhanced_apache`) and Minify are appropriately left alone — they write to subdir `.htaccess` files and were not the total-site outage cause.

### QA probe — safe

- `htaccess_multiviews` is not asserted anywhere in the QA JS suite (`qa/lib/disk-enhanced.js`); dropping the dead probe field is fine.

### Testing

- Travis CI (branch + PR): pass
- PHPCS on changed files: pass
- `tests/test-apache-cache-uri-path.php`: 4/4 pass
- Rule generation spot-check on a local WP install with page cache enabled: core block no longer contains `Options -MultiViews`; still has `RewriteEngine On` and `W3TC_URI_PATH_SLASH`; cache-dir block retains guarded directive

Could not reproduce the full `AllowOverride FileInfo` Apache 500 locally (needs a restrictive vhost), but the repro steps in the PR description match Apache's documented behavior.

### Non-blocking follow-ups

- Existing 2.10.0 installs will need a settings save (or manual `.htaccess` edit) to regenerate rules — worth a release-note callout.
- A small standalone regression test (similar to `tests/test-apache-cache-uri-path.php`) asserting the core block never emits `Options -MultiViews` would guard against reintroduction, but isn't required to merge this hotfix.

PR #1356, opened 2026-07-09
Fix flock(LOCK_UN) called after fclose() in Cache_File

### LINKED ISSUE #499: PHP Fatal error:  Uncaught TypeError: flock(): supplied resource is not a valid stream resource
I'm receiving a fatal php error when file locking is enabled, on every cache clear or post/page modification. The error is this:
PHP Fatal error:  Uncaught TypeError: flock(): supplied resource is not a valid stream resource in /var/www/wp-content/plugins/w3-total-cache/Cache_File.php:246 [....]

This is happening to me with php 8.1, in the previous webserver whith the same settings (file locking on), but with php 7.4 it did not happened.

Could this be an error because the cache file is unlocked after it is closed? There are 3 places where locks are released on that file (Cache_File.php), in two of them the lock is released after the cache file is closed, and on the other it is unlocked before the fclose.

https://github.com/W3EDGE/w3-total-cache/blob/7af6d749297ce6265b6dc50d808fc773c66b1385/Cache_File.php#L246

mosandlt:
## Summary

`Cache_File::set()` and `Cache_File::delete()` call `@flock( $fp, LOCK_UN )` **after** `@fclose( $fp )`. Since PHP 8, calling `flock()` on an already-closed stream resource throws an uncaught `TypeError` ("supplied resource is not a valid stream resource"), which is a fatal error and is **not** suppressible with the `@` operator.

This only triggers when `pgcache.file.locking` is enabled (off by default), which is presumably why it's gone unnoticed - but once enabled, it fires on essentially every single page cache write.

Ran into this on a production site after enabling `pgcache.file.locking` to fix a torn/corrupted page cache file issue (concurrent PHP-FPM workers writing the same large page without locking). The fatal error happens inside the `ob_callback` after page content has already been sent to the client, so it's invisible to site visitors - but it means the disk page cache silently never gets persisted once locking is on, defeating the purpose of turning it on in the first place.

This is exactly the root cause already diagnosed by @mavas84 in #499, which was closed without a code change:

> `flock( $fp, LOCK_UN );` should be called before `fclose( $fp );`

This PR does exactly that, in both places in `Cache_File.php` where the bug occurs. `Cache_File::get()` already has the correct order elsewhere in the same file, so this brings `set()` and `delete()` in line with it.

## Test plan

- [x] `php -l Cache_File.php` - no syntax errors
- [x] Deployed the fix to a live WordPress site with `pgcache.file.locking` enabled; confirmed the previously constant `flock(): supplied resource is not a valid stream resource` fatal errors stopped appearing in the PHP-FPM error log across multiple subsequent page loads, and `x-cache-status: HIT` is now returned correctly on repeat requests

Fixes #499

### DIFF
--- a/Cache_File.php
+++ b/Cache_File.php
@@ -161,11 +161,12 @@ class Cache_File extends Cache_Base {
 		@fputs( $fp, pack( 'L', $expires_at ) );
 		@fputs( $fp, '<?php exit; ?>' );
 		@fputs( $fp, @serialize( $content ) );
-		@fclose( $fp );
 
 		if ( $this->_locking ) {
 			@flock( $fp, LOCK_UN );
 		}
+
+		@fclose( $fp );
 
 		return true;
 	}
@@ -313,11 +314,12 @@ class Cache_File extends Cache_Base {
 				}
 
 				@fputs( $fp, pack( 'L', 0 ) ); // make it expired.
-				@fclose( $fp );
 
 				if ( $this->_locking ) {
 					@flock( $fp, LOCK_UN );
 				}
+
+				@fclose( $fp );
 
 				return true;
 			}


cssjoe (commented):
## Review summary

Thanks for the clear diagnosis and minimal fix — this matches the maintainer guidance in #499 and the correct pattern already used in `Cache_File::_get_with_old_raw()`.

**Verified locally:** with `locking=true`, `Cache_File::set()` on PHP 8.3 reproduces the fatal `TypeError: flock(): supplied resource is not a valid stream resource`. Reordering `flock(LOCK_UN)` before `fclose()` in `set()` and `delete()` resolves it; round-trip `get()` and soft `delete()` also succeed.

**Scope gap:** default page cache uses `pgcache.engine = file_generic` (`Cache_File_Generic`), which overrides `set()` and has the same `fclose()`-before-`flock()` bug in `set()` and `_read()`. Sites on Disk: Enhanced with `pgcache.file.locking` enabled would still hit the fatal after this PR alone.

I'm opening a follow-up PR that fixes `Cache_File_Generic` and adds `tests/test-cache-file-locking.php` covering both backends.

**Recommendation:** Approve and merge this PR. Land the Generic follow-up alongside or immediately after.

cssjoe (approved):

PR #1357, opened 2026-07-09
Fix flock(LOCK_UN) order in Cache_File_Generic and add locking regression test

### LINKED ISSUE #499: PHP Fatal error:  Uncaught TypeError: flock(): supplied resource is not a valid stream resource

cssjoe:
## Summary

- Reorder `flock(LOCK_UN)` before `fclose()` in `Cache_File_Generic::set()` and `Cache_File_Generic::_read()` — the default page cache engine (`pgcache.engine = file_generic`) overrides `set()` and was still calling `flock()` on a closed stream when `pgcache.file.locking` is enabled.
- Add `tests/test-cache-file-locking.php`, a standalone regression test covering both `Cache_File` and `Cache_File_Generic` with `locking=true` on PHP 8+.
- Include the same `Cache_File.php` reorder from #1356 so the regression test passes on one branch; if #1356 merges first, the `Cache_File.php` hunk should be a no-op on rebase.

Fixes #499

Related: #1356

## Test plan

- [x] `php tests/test-cache-file-locking.php` — all 5 checks pass on PHP 8.3
- [x] `php -l Cache_File_Generic.php` — no syntax errors
- [x] `php -l tests/test-cache-file-locking.php` — no syntax errors

### DIFF
--- a/Cache_File.php
+++ b/Cache_File.php
@@ -163,6 +163,7 @@ class Cache_File extends Cache_Base {
 		@fputs( $fp, @serialize( $content ) );
 
 		if ( $this->_locking ) {
+			@\fflush( $fp );
 			@flock( $fp, LOCK_UN );
 		}
 
@@ -316,6 +317,7 @@ class Cache_File extends Cache_Base {
 				@fputs( $fp, pack( 'L', 0 ) ); // make it expired.
 
 				if ( $this->_locking ) {
+					@\fflush( $fp );
 					@flock( $fp, LOCK_UN );
 				}
 
--- a/Cache_File.php
+++ b/Cache_File.php
@@ -161,11 +161,12 @@ class Cache_File extends Cache_Base {
 		@fputs( $fp, pack( 'L', $expires_at ) );
 		@fputs( $fp, '<?php exit; ?>' );
 		@fputs( $fp, @serialize( $content ) );
-		@fclose( $fp );
 
 		if ( $this->_locking ) {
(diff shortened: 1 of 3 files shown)


jacobd91 (dismissed):

cssjoe:
Addressed the Copilot test feedback in 97b2de64:

- moved the standalone cache fixture to a unique `sys_get_temp_dir()` path
- added a PHP < 8 skip for the TypeError-dependent check
- made cache directory creation fail with a clear message
- guarded the `Cache_File_Generic` content assertion before checking it

Verified with:

- `php tests/test-cache-file-locking.php`
- `php -l tests/test-cache-file-locking.php`

jacobd91 (approved):

PR #1358, opened 2026-07-14
Remove Apache Options MultiViews

cssjoe:
To test:

- Test on a server using Apache
- Enable Page Cache and Minify
- Save the settings
- Inspect the `.htaccess` files created.
- Browse the site pages in Incognito

### DIFF
--- a/Minify_Environment.php
+++ b/Minify_Environment.php
@@ -659,9 +659,14 @@ class Minify_Environment {
 		$cache_control = ( $browsercache && $w3tc_config->get_boolean( 'browsercache.cssjs.cache.control' ) );
 		$etag          = ( $browsercache && $w3tc_config->get_integer( 'browsercache.html.etag' ) );
 		$w3tc          = ( $browsercache && $w3tc_config->get_integer( 'browsercache.cssjs.w3tc' ) );
+		$compatibility = $w3tc_config->get_boolean( 'pgcache.compatibility' );
 
 		$rules  = '';
 		$rules .= W3TC_MARKER_BEGIN_MINIFY_CACHE . "\n";
+		// workaround for .gzip.
+		if ( $compatibility ) {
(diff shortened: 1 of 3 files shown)


Minify_Environment.php
```diff
@@ -659,14 +659,9 @@ private function rules_cache_generate_apache( $w3tc_config ) {
 		$cache_control = ( $browsercache && $w3tc_config->get_boolean( 'browsercache.cssjs.cache.control' ) );
 		$etag          = ( $browsercache && $w3tc_config->get_integer( 'browsercache.html.etag' ) );
 		$w3tc          = ( $browsercache && $w3tc_config->get_integer( 'browsercache.cssjs.w3tc' ) );
-		$compatibility = $w3tc_config->get_boolean( 'pgcache.compatibility' );
 
 		$rules  = '';
 		$rules .= W3TC_MARKER_BEGIN_MINIFY_CACHE . "\n";
-		// workaround for .gzip.
```
jacobd91, line 666: Restore these lines and the variable above so it matches pre 2.10.0 behavior

PgCache_Environment.php
```diff
@@ -1681,8 +1681,6 @@ private function rules_cache_generate_apache( $w3tc_config ) {
 		$rules  = '';
 		$rules .= W3TC_MARKER_BEGIN_PGCACHE_CACHE . "\n";
 		if ( $compatibility || $disk_enhanced_apache ) {
-			$rules .= "Options -MultiViews\n";
```
jacobd91, line 1684: Restore this line and remove the $disk_enhanced_apache variable above so it matches pre 2.10.0 behavior (only check for $compatibility)

jacobd91 (changes requested):

jacobd91 (approved):

PR #1363, opened 2026-07-20
Correct placeholder name for string replacement injecting the timeout config

DanWin:
The JS file has recently been re-written, introducing spaces that break the string replacement inside the PHP script. This fixes it.

### DIFF
--- a/UserExperience_DeferScripts_Script.js
+++ b/UserExperience_DeferScripts_Script.js
@@ -1,5 +1,5 @@
 (function (w, k, o) {
-  var t = setTimeout(f, { config_timeout });
+  var t = setTimeout(f, {config_timeout});
   k.forEach(function (e) {
     w.addEventListener(e, f, o);
   });


cssjoe (changes requested):
Thanks for catching this — the regression is real and easy to miss.

## Verified

Regression introduced in Release 2.10.0 (#1344) when `UserExperience_DeferScripts_Script.js` was reformatted: the placeholder became `{ config_timeout }` (spaces), while `UserExperience_DeferScripts_Extension::embed_script()` still does `str_replace( '{config_timeout}', ... )`.

Local checks:
- **master**: replacement is a no-op → served JS keeps `{ config_timeout }` → invalid `setTimeout` argument (`node --check` fails).
- **this PR**: replacement injects the timeout integer correctly → `setTimeout(f, 2500)` and `node --check` passes.
- Travis CI on the PR is green.

## Requested change

Please make this durable against Prettier. With the PR applied, `yarn run js-lint` / `prettier-eslint` immediately rewrites the line back to `{ config_timeout }`, which would re-break the PHP replacement. `js-lint` is currently commented out in `.travis.yml`, so CI does not catch that today, but the next lint-fix or enabling of `js-lint` would silently undo this fix.

Preferred approach: keep the Prettier-formatted JS and update the PHP matcher to accept the spaced form (and optionally both), e.g. in `embed_script()`:

```php
$content = str_replace(
\tarray( '{config_timeout}', '{ config_timeout }' ),
\t$config_timeout,
\t$content
);
```

That way the shipped JS can stay Prettier-clean and the inject keeps working. Alternatively, switch to a non-object-looking token (e.g. `__W3TC_CONFIG_TIMEOUT__`) in both files so Prettier will not treat it as an object literal.

Happy to re-review once that lands.

cssjoe:
Thanks for the report and the fix, @DanWin — the diagnosis is correct (the timeout inject broke after the 2.10.0 Prettier reformat of `UserExperience_DeferScripts_Script.js`).

We're closing this PR in favor of https://github.com/BoldGrid/w3-total-cache/pull/1365, which updates the PHP `str_replace` to match Prettier's spaced placeholder (`{ config_timeout }`) instead of changing the JS.

**Why PHP instead of JS:** removing the spaces in the JS file works until the next `prettier-eslint` / `yarn run js-lint-fix` pass, which immediately rewrites the placeholder back to `{ config_timeout }` and would silently re-break the inject. Matching Prettier's format in PHP keeps lint clean and keeps the fix durable.

Appreciate the contribution — the PHP-side PR lands the same functional fix with that durability in mind.

PR #1365, opened 2026-07-20
ENG7-4410: Match Defer Scripts timeout placeholder to Prettier format

cssjoe:
## Summary
- Fixes Defer Scripts timeout injection broken since 2.10.0 (#1344), when Prettier reformatted `{config_timeout}` to `{ config_timeout }` in `UserExperience_DeferScripts_Script.js` while PHP still replaced the unspaced token.
- Updates `UserExperience_DeferScripts_Extension::embed_script()` to replace `{ config_timeout }` so the configured timeout is injected and the footer script remains valid JS.
- Prefers a PHP-side match over changing the JS (as in #1363) so `prettier-eslint` / `yarn run js-lint` stays clean and a future lint-fix cannot silently re-break the inject.

Closes community approach in favor of this durable fix; see #1363.

Jira: https://imh-internal.atlassian.net/browse/ENG7-4410

## Test plan
- [ ] Confirm `UserExperience_DeferScripts_Script.js` still contains `setTimeout(f, { config_timeout });` (Prettier form)
- [ ] Enable Defer Scripts with a known timeout (e.g. 2500); view page source and confirm the footer script has `setTimeout(f, 2500)` (or the configured value), not a leftover `{ config_timeout }` placeholder
- [ ] `php -r` / `str_replace( '{ config_timeout }', … )` against the JS file succeeds; `node --check` on the injected output passes
- [ ] `yarn run prettier-eslint UserExperience_DeferScripts_Script.js --list-different` exits 0 (no JS change required)

### DIFF
--- a/UserExperience_DeferScripts_Extension.php
+++ b/UserExperience_DeferScripts_Extension.php
@@ -194,7 +194,7 @@ class UserExperience_DeferScripts_Extension {
 
 		$content = file_get_contents( W3TC_DIR . '/UserExperience_DeferScripts_Script.js' );
 		$content = str_replace(
-			'{config_timeout}',
+			'{ config_timeout }',
 			$config_timeout,
 			$content
 		);


jacobd91 (approved):
## Code review + acceptance

Correct, durable fix for the 2.10.0 Prettier mismatch. Preferring the PHP-side token match over changing the JS (vs #1363) is the right call so lint cannot silently re-break injection.

### Verified
- `UserExperience_DeferScripts_Script.js` line 2 is `setTimeout(f, { config_timeout });` (Prettier/spaced form only; no unspaced `{config_timeout}`)
- Old `str_replace( '{config_timeout}', … )` leaves the placeholder in place
- New `str_replace( '{ config_timeout }', 2500, … )` yields `setTimeout(f, 2500)`
- Exercised private `embed_script()` via reflection with timeout `2500`: footer script injects correctly, no leftover placeholder; `node --check` on embedded output passes
- `php -l` clean on the changed PHP file
- Confirmed history: 2.9.4 had `{config_timeout}`; 2.10.0 (#1344) Prettier-expanded to `{ config_timeout }`

### Note
Unreplaced `{ config_timeout }` is still **valid JS** (object shorthand), so syntax checks alone can miss this. Runtime impact is a bad `setTimeout` delay (object → `NaN` → effectively `0`), i.e. scripts fire immediately instead of after the configured timeout — matching the regression report.

Could not run `yarn run prettier-eslint` here (script/deps not available in this checkout); content already matches Prettier form and this PR does not touch the JS.

**Approve.**

jacobd91 (approved):

PR #1381, opened 2026-08-20
Page Cache: add w3tc_pgcache_rules_required filter

tkalimullin-cloudlinux:
# Page Cache: add `w3tc_pgcache_rules_required` filter

Lets an integration tell W3TC not to write its page-cache rewrite rules, while W3TC keeps generating the cache files as usual. With nothing hooked, behaviour is unchanged.

## Why

When Page Cache runs on `file_generic` or `nginx_memcached`, W3TC always owns the rewrite rules, and there is no supported way to ask it to stand down while it keeps building the cache. An integration that serves `page_enhanced/` itself has to either fight W3TC over `.htaccess` / `nginx.conf` or turn page caching off, losing cache generation with it.

The closest thing available today is to inject a condition that can never match through `w3tc_pagecache_rules_apache_rewrite_cond`. That leaves the rules in the file, so they are still parsed and evaluated on every request, and it uses a filter meant for something else.

## The filter

```php
// PgCache_Environment::is_rules_required()
return $required && (bool) \apply_filters( 'w3tc_pgcache_rules_required', $required, $w3tc_c );

// integrator side
add_filter( 'w3tc_pgcache_rules_required', '__return_false' );
```

Both existing callers inherit it, so returning false stops the rules being written and removes blocks written earlier, on the Apache and the nginx path, and `get_required_rules()` returns null so the Install tab stops listing rules that are not on disk. Nothing on the cache-generation path reads this method.

It can only suppress: behind `$required &&` it is not consulted for configurations that never use rewrite rules, so a callback cannot switch rules on where W3TC would not have written them.

Minify already separates writing rewrite rules from generating files, through `minify.rewrite`; page cache has no equivalent. The filter name follows `w3tc_pgcache_rules_apache_last` in the same file.

## Two changes that come with it

- `PgCache_ContentGrabber::_check_rules_present()` returns early when rules are not required. It restores a missing cache-dir `.htaccess` on a front-end request, and once the rules are suppressed that file is never written, so without the exit the restore would fire on every cache miss. It asks through the existing public `get_required_rules()`, whose null return is its cheap branch, so no visibility changes anywhere. The check sits after the existing `file_exists()` exit, so the common path is untouched.
- The `pgcache.debug` rewrite test is gated on the same decision, since otherwise it reports a failure for a rewrite that is deliberately absent. The permalink check next to it still runs.

## Verification

- `phpcs --standard=phpcs.xml` and `php -l` clean on both changed files.
- New standalone test, in the style of `tests/test-apache-cache-rules.php`, with 10 assertions: defaults for `file_generic`, `nginx_memcached`, `file` and page cache disabled; suppression; the config object reaching the callback; the boolean cast; and that a callback cannot force rules on.
  Reverting the filter in the source fails 2 of them, so the test exercises the filter rather than merely running.
- Live stand, Disk: Enhanced. Default: rules and cache-dir `.htaccess` written as before. Suppressed: the rules block is gone from `.htaccess`, cache files keep being written, a front-end request does not bring the rules back, and removing the callback restores the previous state.
- nginx path: both rule blocks in `nginx.conf` removed and restored the same way.
- Restore path instrumented over three forced cache misses: entered on all three; `fix_on_wpadmin_request()` ran three times without the new guard and zero times with it.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 7f15bfb89a7ac947c756b7f030a99e44154270a5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,11 @@ Run the standalone cache-unserialize helper test (no WordPress bootstrap needed)
 php tests/test-cache-unserialize.php
 ```
 
+Run the standalone page-cache rules-required filter test (no WordPress bootstrap needed):
+```bash
+php tests/test-pgcache-rules-required-filter.php
+```
+
 #### Writing new standalone tests
 
 Standalone test files in `tests/` use a realpath-guard at the top so they exit early when PHPUnit auto-discovers them, e.g.:
--- a/PgCache_ContentGrabber.php
+++ b/PgCache_ContentGrabber.php
@@ -1313,8 +1313,15 @@ class PgCache_ContentGrabber {
 			return;
 		}
 
-		// we call it as little times as possible its expensive, but have to restore lost .htaccess file.
 		$e = Dispatcher::component( 'PgCache_Environment' );
+
+		// When the "w3tc_pgcache_rules_required" filter suppresses the rules this file is never
+		// written, so without the exit the restore below would repeat on every cache miss.
+		if ( null === $e->get_required_rules( $this->_config ) ) {
+			return;
+		}
+
+		// we call it as little times as possible its expensive, but have to restore lost .htaccess file.
 		try {
 			$e->fix_on_wpadmin_request( $this->_config, true );
 		} catch ( \Exception $ex ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedCatch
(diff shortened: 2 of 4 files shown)


cssjoe (approved):
@tkalimullin-cloudlinux Thank you for your contribution.  We'll include this in our next release.

PR #1383, opened 2026-08-31
ENG3-590: Improve minify URLs, Java tools, TLS defaults, diagnostic logs, and proxy headers

jacobd91:
## Summary
- Validate minify precache URLs before outbound fetch
- Apply host allowlists to constant URL overrides
- Validate Java minifier JAR paths before invocation
- Enable verified TLS defaults for BunnyCDN and Redis
- Extend diagnostic log redaction across remaining formats
- Validate proxy-derived client IP and scheme headers

## Behavior changes to watch
- TLS verification defaults to on for Redis (`tls:` stream contexts) and Bunny CDN API calls. Bunny CDN and CDN FSD each gain a **Verify TLS Certificates** setting for compatibility opt-out, wired through the purge and AJAX flows.
- Forwarded headers (`X-Forwarded-For`, `X-Real-IP`, `X-Forwarded-Proto`, `Forwarded`, `CF-Connecting-IP`) are applied only when the TCP peer is listed in `common.trusted_proxies`, plus Cloudflare ranges when that extension is active.
- Constant URL overrides (license, notices, PageSpeed, API2, TransparentCDN) must be HTTPS on their allowlisted hosts; anything else falls back to the built-in default.
- Minify precache and `Util_Http::download()` check each URL and each redirect hop with `Util_Url::is_allowed_outbound_url()`; redirects are followed manually with `redirection => 0`.
- Java-backed minifiers (`yuijs`, `yuicss`, `ccjs`) run both the Java binary and the JAR path through `Util_Java::validate_tools()` as part of the `available()` check.
- `Util_Debug::redact()` covers more diagnostic formats (query strings, JSON, `print_r`, cookies, URIs, headers) and is applied to module logs, audit events, and support attachments.

## Test plan
- [x] targeted PHPUnit: 137 tests, 421 assertions
- [x] full automated checks in CI

Internal tracking: [ENG3-590](https://imh-internal.atlassian.net/browse/ENG3-590)

[ENG3-590]: https://imh-internal.atlassian.net/browse/ENG3-590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cssjoe (changes requested):
Requesting changes — two items to resolve before merge, both about compatibility on existing installs rather than the approach itself.

For context on the rest: I ran the touched PHPUnit classes locally against `be83c64` and they pass (137 tests, 421 assertions), and phpcs, semgrep, conftest, and Travis are green.

### 1. `common.trusted_proxies` has no UI and no upgrade path

`Util_Environment::is_https()` and `get_client_ip()` now apply forwarded headers only when the peer address falls inside `common.trusted_proxies` (`Util_Environment.php` around L2199; key added in `ConfigKeys.php` around L2437). That key defaults to an empty array and there is no admin field for it, so an install behind a TLS-terminating load balancer that relies on `X-Forwarded-Proto` today will begin reporting HTTP after upgrade. That flows into `pgcache.cache.ssl` handling, protocol-relative minify URLs, and CDN scheme selection on sites that currently work.

Please either add a General Settings field for the CIDR list, or document the config key and the `w3tc_trusted_proxies` filter in the upgrade notice, so operators can set it before it becomes required.

Related test gap: `tests/admin/class-w3tc-proxy-headers-test.php` only drives the `w3tc_trusted_proxies` filter. Nothing asserts the config-array path (`$config->set( 'common.trusted_proxies', … )`), which is the path a real site would use. Please add coverage for it so the config plumbing cannot regress silently.

### 2. Bunny purge reads the TLS flag from the wrong surface

`Cdn_BunnyCdn_Page.php` L158–L160 chooses `cdn.bunnycdn.verify_tls_certificates` whenever `cdn.bunnycdn.pull_zone_id` is non-empty, and the `cdnfsd` key otherwise. On a full-site-delivery install that still has a leftover object-CDN pull zone id stored, the FSD opt-out is ignored and the purge uses the CDN flag instead.

Please resolve the flag from the surface actually being purged (or require both to be enabled), and add a case covering an FSD opt-out alongside a stale CDN zone id.

Rest of the change set reads well — the `validate_tools()` gate in `available()` (rather than only `init()`) and the manual redirect loop in `Util_Http::download()` are both the right call. Notes and follow-ups are on ENG3-590.

jacobd91:
Addressed both review items on `3001e74`:

1. **Trusted reverse proxies** — General Settings → Reverse Proxy now has a CIDR/IP textarea (`common.trusted_proxies`). Empty still ignores forwarded headers. Tests cover the config-array path as well as the existing filter.
2. **Bunny URL-purge TLS** — the flag follows the enabled Bunny surface(s). A leftover object-CDN pull zone id no longer overrides an FSD opt-out. When both CDN and FSD Bunny engines are enabled, both flags must be on.

Targeted PHPUnit: 142 tests, 434 assertions.

cssjoe (approved):
Approved on `3001e74`.

Both items from the previous review are addressed:

1. General Settings → Reverse Proxy now has the trusted reverse-proxy CIDR/IP field (`common.trusted_proxies`). Tests cover the config-array path as well as the filter.
2. Bunny URL-purge TLS follows the enabled Bunny surface(s). A leftover object-CDN pull zone id no longer overrides an FSD opt-out. When both CDN and FSD Bunny engines are enabled, both flags must be on.

Targeted PHPUnit: 142 tests, 434 assertions. phpcs, semgrep, conftest, and Travis are green.

PR #1387, opened 2026-09-03
ENG3-781: Fix Google PageSpeed access token auto-renew

jacobd91:
https://imh-internal.atlassian.net/browse/ENG3-781

## Summary

Google PageSpeed authorization did not reliably auto-renew. Once the stored access token expired, users were sent back to the **Authorize** button on General Settings — and in some cases the stored PageSpeed credentials were discarded entirely — even when the account was still authorized and renewal should have succeeded.

`PageSpeed_Api::refresh_token()` had three gaps that made a failed renewal indistinguishable from a successful one:

1. **HTTP status was never inspected.** Only a nested `error.code` was treated as a failure, so a token endpoint that answers with a Google error payload under an HTTP `200` looked like a success with no access token. Transport-level `WP_Error` responses were returned from the method and then ignored by the caller.
2. **Credentials were cleared too aggressively.** Any `refresh-token-not-found` / `refresh-token-missing-refresh-token` response wiped both `widget.pagespeed.access_token` and `widget.pagespeed.w3tc_pagespeed_key`, forcing a full re-authorization even when the failure was transient.
3. **Renewed tokens could re-expire immediately.** A response without a `created` timestamp was stored as-is, so `isAccessTokenExpired()` stayed `true` and the widget kept trying to refresh.

## Changes

- Fail closed on a non-2xx status, a nested W3 API `error.id`, **or** a flat Google `error` string, so an error body is never mistaken for a renewed token.
- Handle `WP_Error` from the request explicitly and surface the transport message in the admin notice.
- Validate that a successful response carries a string `access_token` and a numeric `expires_in` before persisting it.
- Default `created` to the current time when the response omits it, so a freshly renewed token is not treated as already expired.
- Only clear the stored access token and API key for irrecoverable responses (`refresh-token-not-found`, `refresh-token-missing-refresh-token`). Transport errors, 5xx responses, and Google rejections now keep the stored credentials so the next attempt can succeed.
- Extract `set_refresh_failure()`, `get_refresh_failure_message()`, and `clear_pagespeed_credentials()` helpers, and add operator-facing text for each documented error id.

A companion server-side change to the W3 API service is required for the full fix: the token endpoint must report Google refresh failures with a non-2xx status and persist rotated refresh tokens. This PR is the plugin half and is safe to merge independently — it stops the plugin from discarding good credentials or storing an already-expired token.

## Test plan

New coverage in `tests/admin/class-w3tc-pagespeed-refresh-token-test.php` (5 tests, 16 assertions) using an in-memory config stub and a mocked `pre_http_request`:

- A successful renewal persists the token to config, loads it into the client, adds a `created` timestamp, and leaves `isAccessTokenExpired()` false.
- A flat Google `{"error":"invalid_grant"}` body returned with HTTP 200 is treated as a failure and keeps the stored credentials.
- An HTTP 502 service error keeps the stored credentials.
- An HTTP 428 `refresh-token-not-found` clears the stored token and key.
- A transport `WP_Error` keeps the stored credentials and records the failure message for the admin notice.

```
$ phpunit --filter W3tc_Pagespeed_Refresh_Token_Test
OK (5 tests, 16 assertions)
```

`php -l` and `phpcs` are clean on the changed files.

Manual verification of an end-to-end renewal against a patched API service is pending and tracked as acceptance testing.

## Live authorization follow-up

The first dev authorization attempt exposed a distinct missing-refresh-token response from the companion API. This PR now maps `authorize-out-refresh-token-missing` to a clear admin message instructing the user to authorize again. The API companion PR also switches Google OAuth to `prompt=consent` and corrects the encoded error callback URL.

---

> [!NOTE]
> **Medium Risk**
> Changes OAuth token persistence, automatic refresh, and credential clearing in admin-facing Google PageSpeed flows; behavior is improved and tested but still touches security-sensitive authorization state.
> 
> **Overview**
> Fixes unreliable Google PageSpeed **access token auto-renew** by making `PageSpeed_Api::refresh_token()` fail closed on non-2xx responses, nested W3 API errors, flat Google `error` payloads, and `WP_Error` transport failures—instead of treating error bodies as successful renewals.
> 
> **Refresh behavior** now validates `access_token` / `expires_in`, stamps `created` on save so renewed tokens are not immediately expired, backs off automatic retries for **15 minutes** via `w3tcps_refresh_retry_after`, and only clears stored credentials for irrecoverable API IDs (`refresh-token-not-found`, `refresh-token-missing-refresh-token`). Transient failures keep local OAuth state and surface clearer admin notices through new helpers (`set_refresh_failure`, `get_refresh_failure_message`, `clear_pagespeed_credentials`, `clear_failure_notices`).
> 
> **Revoke / authorize** paths are aligned: `reset()` returns a bool, uses HTTP-aware error mapping, treats `revoke-token-not-found` as success with local cleanup, and authorization errors in `general.php` go through `get_authorize_failure_message()` (including new refresh-token-missing cases). Successful authorize clears prior failure options.
> 
> **PageSpeed page and dashboard widget** AJAX now treats an **expired** token like missing when live data is needed, while **cached** results still render without outbound calls.
> 
> Adds PHPUnit coverage for refresh/revoke edge cases and widget cache vs. expiry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd839a3a177a7e3da211807ea67bc84068055789. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/PageSpeed_Api.php
+++ b/PageSpeed_Api.php
@@ -18,6 +18,13 @@ namespace W3TC;
  */
 class PageSpeed_Api {
 	/**
+	 * Delay automatic retries after a failed token refresh.
+	 *
+	 * @var int
+	 */
+	const REFRESH_RETRY_DELAY = 15 * MINUTE_IN_SECONDS;
+
+	/**
 	 * Config.
 	 *
 	 * @var object
@@ -230,7 +237,13 @@ class PageSpeed_Api {
 	public function maybe_refresh_token() {
 		$w3tc_site_id       = Util_Http::generate_site_id();
 		$w3tc_pagespeed_key = $this->w3tc_config->get_string( 'widget.pagespeed.w3tc_pagespeed_key' );
-		if ( $this->client->isAccessTokenExpired() && ! empty( $w3tc_pagespeed_key ) ) {
+		$retry_after        = (int) get_option( 'w3tcps_refresh_retry_after' );
+
+		if (
+			$this->client->isAccessTokenExpired() &&
+			! empty( $w3tc_pagespeed_key ) &&
+			$retry_after <= time()
+		) {
 			$this->refresh_token( $w3tc_site_id, $w3tc_pagespeed_key );
 		}
 	}
@@ -324,6 +337,7 @@ class PageSpeed_Api {
 		$this->client->setAccessToken( $w3tc_access_token );
 		delete_option( 'w3tcps_refresh_fail' );
 		delete_option( 'w3tcps_refresh_fail_message' );
+		delete_option( 'w3tcps_refresh_retry_after' );
 	}
 
 	/**
@@ -339,6 +353,7 @@ class PageSpeed_Api {
 	private function set_refresh_failure( $title, $message = '' ) {
 		update_option( 'w3tcps_refresh_fail', $title );
 		update_option( 'w3tcps_refresh_fail_message', $message );
+		update_option( 'w3tcps_refresh_retry_after', time() + self::REFRESH_RETRY_DELAY );
 	}
 
 	/**
@@ -368,6 +383,9 @@ class PageSpeed_Api {
 
 			case 'refresh-token-google-failed':
 				return __( 'Google rejected the access token refresh. Local authorization was kept for retry.', 'w3-total-cache' );
+
+			case 'refresh-token-persist-failed':
+				return __( 'The renewed Google access token could not be saved by the W3 API. Local authorization was kept for retry.', 'w3-total-cache' );
 		}
 
 		if ( isset( $response_body['error'] ) && is_string( $response_body['error'] ) ) {
@@ -464,9 +482,66 @@ class PageSpeed_Api {
 	}
 
 	/**
+	 * Returns operator-facing text for an authorization failure.
+	 *
+	 * @since X.X.X
+	 *
+	 * @param string $error_json W3 API authorization error JSON.
+	 *
+	 * @return string
+	 */
+	public static function get_authorize_failure_message( $error_json ) {
+		$error    = json_decode( $error_json );
+		$error_id = isset( $error->error->id ) ? (string) $error->error->id : '';
+
+		switch ( $error_id ) {
+			case 'authorize-in-missing-site-id':
+				return __( 'Unique site ID missing for authorize request!', 'w3-total-cache' );
+
+			case 'authorize-in-missing-auth-url':
+				return __( 'Authorize URL missing for authorize request!', 'w3-total-cache' );
+
+			case 'authorize-in-missing-return-url':
+				return __( 'Return URL missing for authorize request!', 'w3-total-cache' );
+
+			case 'authorize-in-failed':
+				return __( 'Failed to process authorize request!', 'w3-total-cache' );
+
+			case 'authorize-out-code-missing':
+				return __( 'No authorize code returned to W3-API from Google!', 'w3-total-cache' );
+
+			case 'authorize-out-site-id-missing':
+				return __( 'Unique site ID missing during Google authorization return processing!', 'w3-total-cache' );
+
+			case 'authorize-out-w3tc-pagespeed-key-missing':
+				return __( 'No W3Key returned to W3-API from Google!', 'w3-total-cache' );
+
+			case 'authorize-out-not-found':
+				return __( 'No W3-API matching record found during Google authorization return processing!', 'w3-total-cache' );
+
+			case 'authorize-out-token-missing':
+				return __( 'No Google access token found during Google authorization return processing!', 'w3-total-cache' );
+
+			case 'authorize-out-refresh-token-missing':
+				return __( 'Google did not return a refresh token. Please authorize Google PageSpeed again.', 'w3-total-cache' );
+
+			case 'authorize-out-refresh-token-persist-failed':
+				return __( 'The Google refresh token could not be saved by the W3 API. Please try authorizing again after the service is updated.', 'w3-total-cache' );
+		}
+
+		return sprintf(
+			/* translators: %s: Google authorization error ID. */
+			__( 'Unexpected Google PageSpeed authorization error: %s', 'w3-total-cache' ),
+			'' !== $error_id ? $error_id : __( 'unknown', 'w3-total-cache' )
+		);
+	}
+
+	/**
 	 * Reset authentication.
 	 *
 	 * @since 2.3.0
+	 *
+	 * @return bool Whether local and remote authorization were reset.
 	 */
 	public function reset() {
 		$w3tc_access_token  = $this->client->getAccessToken();
@@ -478,7 +553,7 @@ class PageSpeed_Api {
 				'w3tcps_revoke_fail',
 				__( 'Google PageSpeed access token revocation missing required parameters!', 'w3-total-cache' )
 			);
-			return;
+			return false;
 		}
 
 		$request = Util_Environment::get_api_base_url() . '/google/revoke-token/' . rawurlencode( $w3tc_access_token ) . '/' . rawurlencode( $w3tc_site_id ) . '/' . rawurlencode( $w3tc_pagespeed_key );
@@ -490,41 +565,88 @@ class PageSpeed_Api {
 			)
 		);
 
-		$response_body_json = wp_remote_retrieve_body( $response );
-		$response_body      = json_decode( $response_body_json, true );
-
 		if ( is_wp_error( $response ) ) {
-			return wp_json_encode(
-				array(
-					'error' => array(
-						'code'    => $response->get_error_code(),
-						'message' => $response->get_error_message(),
-					),
-				)
+			$this->set_revoke_failure(
+				$response->get_error_message()
 			);
-		} elseif ( isset( $response_body['error']['code'] ) && 200 !== $response_body['error']['code'] ) {
-			if ( 'revoke-token-access-token-missing' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No access token provided for revoke!', 'w3-total-cache' );
-			} elseif ( 'revoke-token-api-key-missing' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No W3TC API key provided for revoke!', 'w3-total-cache' );
-			} elseif ( 'revoke-token-not-found' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No matching Google access record found for W3TC API key!', 'w3-total-cache' );
-			}
+			return false;
+		}
 
-			update_option(
-				'w3tcps_revoke_fail',
-				__( 'Google PageSpeed Access Token revocation failed.', 'w3-total-cache' )
-			);
-			update_option(
-				'w3tcps_revoke_fail_message',
-				$w3tc_message
-			);
+		$response_code = (int) wp_remote_retrieve_response_code( $response );
+		$response_body = json_decode( wp_remote_retrieve_body( $response ), true );
+		if ( ! is_array( $response_body ) ) {
+			$response_body = array();
+		}
 
-			return;
+		$w3tc_error_id = '';
+		if ( isset( $response_body['error'] ) && is_array( $response_body['error'] ) ) {
+			$w3tc_error_id = isset( $response_body['error']['id'] ) ? (string) $response_body['error']['id'] : '';
+		}
+
+		if (
+			$response_code < 200 ||
+			$response_code >= 300 ||
+			isset( $response_body['error'] )
+		) {
+			$this->set_revoke_failure(
+				$this->get_revoke_failure_message( $w3tc_error_id, $response_code )
+			);
+			return false;
 		}
 
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', '' );
 		$this->w3tc_config->set( 'widget.pagespeed.w3tc_pagespeed_key', '' );
 		$this->w3tc_config->save();
+		return true;
+	}
+
+	/**
+	 * Records a token revocation failure for the admin notice.
+	 *
+	 * @since X.X.X
+	 *
+	 * @param string $message Failure detail.
+	 *
+	 * @return void
+	 */
+	private function set_revoke_failure( $message ) {
+		update_option(
+			'w3tcps_revoke_fail',
+			__( 'Google PageSpeed Access Token revocation failed.', 'w3-total-cache' )
+		);
+		update_option( 'w3tcps_revoke_fail_message', $message );
+	}
+
+	/**
+	 * Returns operator-facing text for a token revocation failure.
+	 *
+	 * @since X.X.X
+	 *
+	 * @param string $w3tc_error_id Nested W3 API error ID, if any.
+	 * @param int    $response_code HTTP response code.
+	 *
+	 * @return string
+	 */
+	private function get_revoke_failure_message( $w3tc_error_id, $response_code ) {
+		switch ( $w3tc_error_id ) {
+			case 'revoke-token-access-token-missing':
+				return __( 'No access token provided for revoke!', 'w3-total-cache' );
+
+			case 'revoke-token-site-id-missing':
+				return __( 'No site ID provided for revoke!', 'w3-total-cache' );
+
+			case 'revoke-token-api-key-missing':
+				return __( 'No W3TC API key provided for revoke!', 'w3-total-cache' );
+
+			case 'revoke-token-not-found':
+				return __( 'No matching Google access record found for W3TC API key!', 'w3-total-cache' );
+		}
+
+		return sprintf(
+			/* translators: 1: HTTP response code, 2: API error ID. */
+			__( 'Unexpected token revocation response (HTTP %1$d, error %2$s).', 'w3-total-cache' ),
+			$response_code,
+			'' !== $w3tc_error_id ? $w3tc_error_id : __( 'unknown', 'w3-total-cache' )
+		);
 	}
 }
--- a/PageSpeed_Api.php
+++ b/PageSpeed_Api.php
@@ -316,9 +316,7 @@ class PageSpeed_Api {
 			return;
 		}
 
-		if ( empty( $response_body['created'] ) ) {
-			$response_body['created'] = time();
-		}
+		$response_body['created'] = time();
 
 		$w3tc_access_token = wp_json_encode( $response_body );
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', $w3tc_access_token );
@@ -369,11 +367,11 @@ class PageSpeed_Api {
 				return __( 'Matching Google access record found but the refresh token value is blank!', 'w3-total-cache' );
 
 			case 'refresh-token-google-failed':
-				return __( 'Google rejected the access token refresh. Re-authorize Google PageSpeed if this continues.', 'w3-total-cache' );
+				return __( 'Google rejected the access token refresh. Local authorization was kept for retry.', 'w3-total-cache' );
 		}
 
 		if ( isset( $response_body['error'] ) && is_string( $response_body['error'] ) ) {
-			return __( 'Google rejected the access token refresh. Re-authorize Google PageSpeed if this continues.', 'w3-total-cache' );
+			return __( 'Google rejected the access token refresh. Local authorization was kept for retry.', 'w3-total-cache' );
 		}
 
 		return sprintf(
--- a/PageSpeed_Api.php
+++ b/PageSpeed_Api.php
@@ -245,8 +245,7 @@ class PageSpeed_Api {
 	 */
 	public function refresh_token( $w3tc_site_id, $w3tc_pagespeed_key ) {
 		if ( empty( $w3tc_site_id ) || empty( $w3tc_pagespeed_key ) ) {
-			update_option(
-				'w3tcps_refresh_fail',
+			$this->set_refresh_failure(
 				__( 'Google PageSpeed access token refresh missing required parameters!', 'w3-total-cache' )
 			);
 			return;
@@ -261,59 +260,138 @@ class PageSpeed_Api {
 			)
 		);
 
-		$response_body_json = wp_remote_retrieve_body( $response );
-		$response_body      = json_decode( $response_body_json, true );
-
 		if ( is_wp_error( $response ) ) {
-			return wp_json_encode(
-				array(
-					'error' => array(
-						'code'    => $response->get_error_code(),
-						'message' => $response->get_error_message(),
-					),
-				)
+			$this->set_refresh_failure(
+				__( 'Google PageSpeed access token refresh failed.', 'w3-total-cache' ),
+				$response->get_error_message()
 			);
-		} elseif ( isset( $response_body['error']['code'] ) && 200 !== $response_body['error']['code'] ) {
-			if ( 'refresh-token-missing-site-id' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No site ID provided for access key refresh!', 'w3-total-cache' );
-			} elseif ( 'refresh-token-missing-w3tc-pagespeed-key' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No W3TC API key provided for access key refresh!', 'w3-total-cache' );
-			} elseif ( 'refresh-token-not-found' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No matching Google access record found for W3TC API key!', 'w3-total-cache' );
-			} elseif ( 'refresh-token-missing-refresh-token' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'Matching Google access record found but the refresh token value is blank!', 'w3-total-cache' );
-			}
-
-			update_option(
-				'w3tcps_refresh_fail',
-				__( 'Google PageSpeed access token refresh failed.', 'w3-total-cache' )
-			);
-			update_option(
-				'w3tcps_refresh_fail_message',
-				$w3tc_message
-			);
-
-			// Reset the token and key.
-			$this->w3tc_config->set( 'widget.pagespeed.access_token', '' );
-			$this->w3tc_config->set( 'widget.pagespeed.w3tc_pagespeed_key', '' );
-			$this->w3tc_config->save();
-
 			return;
 		}
 
-		$w3tc_access_token = $response_body_json;
+		$response_code = (int) wp_remote_retrieve_response_code( $response );
+		$response_body = json_decode( wp_remote_retrieve_body( $response ), true );
+		if ( ! is_array( $response_body ) ) {
+			$response_body = array();
+		}
 
-		if ( empty( $w3tc_access_token ) || empty( $response_body['access_token'] ) ) {
-			update_option(
-				'w3tcps_refresh_fail',
+		$w3tc_error_id = '';
+		if ( isset( $response_body['error'] ) && is_array( $response_body['error'] ) ) {
+			$w3tc_error_id = isset( $response_body['error']['id'] ) ? (string) $response_body['error']['id'] : '';
+		}
+
+		if (
+			$response_code < 200 ||
+			$response_code >= 300 ||
+			isset( $response_body['error'] )
+		) {
+			$this->set_refresh_failure(
+				__( 'Google PageSpeed access token refresh failed.', 'w3-total-cache' ),
+				$this->get_refresh_failure_message( $w3tc_error_id, $response_body, $response_code )
+			);
+
+			if (
+				in_array(
+					$w3tc_error_id,
+					array(
+						'refresh-token-not-found',
+						'refresh-token-missing-refresh-token',
+					),
+					true
+				)
+			) {
+				$this->clear_pagespeed_credentials();
+			}
+			return;
+		}
+
+		if (
+			empty( $response_body['access_token'] ) ||
+			! is_string( $response_body['access_token'] ) ||
+			empty( $response_body['expires_in'] ) ||
+			! is_numeric( $response_body['expires_in'] )
+		) {
+			$this->set_refresh_failure(
 				__( 'Google PageSpeed access token refresh failed due to response missing access token.', 'w3-total-cache' )
 			);
 			return;
 		}
 
+		if ( empty( $response_body['created'] ) ) {
+			$response_body['created'] = time();
+		}
+
+		$w3tc_access_token = wp_json_encode( $response_body );
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', $w3tc_access_token );
 		$this->w3tc_config->save();
 		$this->client->setAccessToken( $w3tc_access_token );
+	}
+
+	/**
+	 * Records a token refresh failure for the admin notice.
+	 *
+	 * @since X.X.X
+	 *
+	 * @param string $title   Notice title.
+	 * @param string $message Optional failure detail.
+	 *
+	 * @return void
+	 */
+	private function set_refresh_failure( $title, $message = '' ) {
+		update_option( 'w3tcps_refresh_fail', $title );
+		update_option( 'w3tcps_refresh_fail_message', $message );
+	}
+
+	/**
+	 * Returns operator-facing text for a token refresh failure.
+	 *
+	 * @since X.X.X
+	 *
+	 * @param string $w3tc_error_id Nested W3 API error ID, if any.
+	 * @param array  $response_body Decoded response body.
+	 * @param int    $response_code HTTP response code.
+	 *
+	 * @return string
+	 */
+	private function get_refresh_failure_message( $w3tc_error_id, $response_body, $response_code ) {
+		switch ( $w3tc_error_id ) {
+			case 'refresh-token-missing-site-id':
+				return __( 'No site ID provided for access key refresh!', 'w3-total-cache' );
+
+			case 'refresh-token-missing-w3tc-pagespeed-key':
+				return __( 'No W3TC API key provided for access key refresh!', 'w3-total-cache' );
+
+			case 'refresh-token-not-found':
+				return __( 'No matching Google access record found for W3TC API key!', 'w3-total-cache' );
+
+			case 'refresh-token-missing-refresh-token':
+				return __( 'Matching Google access record found but the refresh token value is blank!', 'w3-total-cache' );
+
+			case 'refresh-token-google-failed':
+				return __( 'Google rejected the access token refresh. Re-authorize Google PageSpeed if this continues.', 'w3-total-cache' );
+		}
+
+		if ( isset( $response_body['error'] ) && is_string( $response_body['error'] ) ) {
+			return __( 'Google rejected the access token refresh. Re-authorize Google PageSpeed if this continues.', 'w3-total-cache' );
+		}
+
+		return sprintf(
+			/* translators: %d: HTTP response code. */
+			__( 'Unexpected token refresh response (HTTP %d).', 'w3-total-cache' ),
+			$response_code
+		);
+	}
+
+	/**
+	 * Removes local PageSpeed credentials after an irrecoverable API response.
+	 *
+	 * @since X.X.X
+	 *
+	 * @return void
+	 */
+	private function clear_pagespeed_credentials() {
+		$this->w3tc_config->set( 'widget.pagespeed.access_token', '' );
+		$this->w3tc_config->set( 'widget.pagespeed.w3tc_pagespeed_key', '' );
+		$this->w3tc_config->save();
 	}
 
 	/**
--- a/PageSpeed_Api.php
+++ b/PageSpeed_Api.php
@@ -324,6 +324,8 @@ class PageSpeed_Api {
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', $w3tc_access_token );
 		$this->w3tc_config->save();
 		$this->client->setAccessToken( $w3tc_access_token );
+		delete_option( 'w3tcps_refresh_fail' );
+		delete_option( 'w3tcps_refresh_fail_message' );
 	}
 
 	/**
--- a/PageSpeed_Api.php
+++ b/PageSpeed_Api.php
@@ -535,6 +535,9 @@ class PageSpeed_Api {
 			case 'authorize-out-token-missing':
 				return __( 'No Google access token found during Google authorization return processing!', 'w3-total-cache' );
 
+			case 'authorize-out-google-failed':
+				return __( 'Google rejected the authorization request. Please try authorizing Google PageSpeed again.', 'w3-total-cache' );
+
 			case 'authorize-out-refresh-token-missing':
 				return __( 'Google did not return a refresh token. Please authorize Google PageSpeed again.', 'w3-total-cache' );
 
--- a/PageSpeed_Api.php
+++ b/PageSpeed_Api.php
@@ -332,12 +332,17 @@ class PageSpeed_Api {
 		$response_body['created'] = time();
 
 		$w3tc_access_token = wp_json_encode( $response_body );
+		if ( false === $w3tc_access_token ) {
+			$this->set_refresh_failure(
+				__( 'Google PageSpeed access token refresh failed because the response could not be encoded.', 'w3-total-cache' )
+			);
+			return;
+		}
+
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', $w3tc_access_token );
 		$this->w3tc_config->save();
 		$this->client->setAccessToken( $w3tc_access_token );
-		delete_option( 'w3tcps_refresh_fail' );
-		delete_option( 'w3tcps_refresh_fail_message' );
-		delete_option( 'w3tcps_refresh_retry_after' );
+		self::clear_failure_notices();
 	}
 
 	/**
@@ -410,6 +415,7 @@ class PageSpeed_Api {
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', '' );
 		$this->w3tc_config->set( 'widget.pagespeed.w3tc_pagespeed_key', '' );
 		$this->w3tc_config->save();
+		delete_option( 'w3tcps_refresh_retry_after' );
 	}
 
 	/**
@@ -491,8 +497,15 @@ class PageSpeed_Api {
 	 * @return string
 	 */
 	public static function get_authorize_failure_message( $error_json ) {
+		if ( ! is_string( $error_json ) ) {
+			$error_json = '';
+		}
+
 		$error    = json_decode( $error_json );
 		$error_id = isset( $error->error->id ) ? (string) $error->error->id : '';
+		if ( 1 !== preg_match( '/\A[a-z0-9-]+\z/', $error_id ) ) {
+			$error_id = '';
+		}
 
 		switch ( $error_id ) {
 			case 'authorize-in-missing-site-id':
@@ -534,6 +547,21 @@ class PageSpeed_Api {
 			__( 'Unexpected Google PageSpeed authorization error: %s', 'w3-total-cache' ),
 			'' !== $error_id ? $error_id : __( 'unknown', 'w3-total-cache' )
 		);
+	}
+
+	/**
+	 * Clears PageSpeed failure notices and automatic refresh cooldown.
+	 *
+	 * @since X.X.X
+	 *
+	 * @return void
+	 */
+	public static function clear_failure_notices() {
+		delete_option( 'w3tcps_refresh_fail' );
+		delete_option( 'w3tcps_refresh_fail_message' );
+		delete_option( 'w3tcps_refresh_retry_after' );
+		delete_option( 'w3tcps_revoke_fail' );
+		delete_option( 'w3tcps_revoke_fail_message' );
 	}
 
 	/**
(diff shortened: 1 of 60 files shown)


PageSpeed_Api.php
```diff
@@ -261,61 +260,140 @@ public function refresh_token( $w3tc_site_id, $w3tc_pagespeed_key ) {
...
 		}
 
+		if ( empty( $response_body['created'] ) ) {
+			$response_body['created'] = time();
+		}
+
+		$w3tc_access_token = wp_json_encode( $response_body );
 		$this->w3tc_config->set( 'widget.pagespeed.access_token', $w3tc_access_token );
 		$this->w3tc_config->save();
 		$this->client->setAccessToken( $w3tc_access_token );
```
cssjoe, line 344: **Blocking:** A successful persist still leaves `w3tcps_refresh_fail` / `w3tcps_refresh_fail_message` in place.

Transient failures now keep credentials *and* record those options. A later successful renew can therefore still show the earlier error on the next admin page, including next to a Valid status on General Settings.

`errors_notice()` only deletes the options after display, so a fail → later success path still surfaces stale copy.

Fix: after a successful persist, `delete_option( 'w3tcps_refresh_fail' )` and `delete_option( 'w3tcps_refresh_fail_message' )`.

Test: seed both options, mock a 200 with `access_token` + `expires_in`, call `refresh_token()`, assert both options are false. Bugbot already drafted this on 23e60c7; please land it here rather than leaving the thread open.

tests/admin/class-w3tc-pagespeed-refresh-token-test.php
```diff
@@ -0,0 +1,299 @@
...
+				),
+			)
+		);
+
+		$api = $this->create_api();
+		$api->refresh_token( 'site-id', 'pagespeed-key' );
+
+		$this->assertSame( '', $this->config->get_string( 'widget.pagespeed.access_token' ) );
+		$this->assertSame( '', $this->config->get_string( 'widget.pagespeed.w3tc_pagespeed_key' ) );
+	}
```
cssjoe, line 415: Please also add a case for `refresh-token-missing-refresh-token` that asserts credentials are cleared. `clear_pagespeed_credentials()` lists two irrecoverable IDs; only `refresh-token-not-found` is covered today, so the sibling wipe path can regress unnoticed.

Non-blocking: `refresh-token-google-failed` keeps credentials (good for retry) but the operator message says to re-authorize. Consider language that matches retryable vs irrecoverable.

PageSpeed_Api.php
```diff
@@ -261,61 +260,140 @@ public function refresh_token( $w3tc_site_id, $w3tc_pagespeed_key ) {
...
+			! is_numeric( $response_body['expires_in'] )
+		) {
+			$this->set_refresh_failure(
 				__( 'Google PageSpeed access token refresh failed due to response missing access token.', 'w3-total-cache' )
 			);
 			return;
 		}
 
+		if ( empty( $response_body['created'] ) ) {
+			$response_body['created'] = time();
```
cssjoe: **Blocking:** `created` is only filled when empty. `W3TCG_Google_Auth_OAuth2::isAccessTokenExpired()` treats a missing or old `created` as expired (`created + expires_in - 30 < time()`).

If the token endpoint returns a new `access_token` but a stale non-empty `created` from the stored record, this branch is skipped, `isAccessTokenExpired()` stays true, and `maybe_refresh_token()` will keep retrying — the original “renewed token looks expired immediately” failure mode.

Fix: on successful refresh, always set `$response_body['created'] = time();` (do not trust a caller-supplied timestamp).

Test: mock a 200 body with `access_token`, `expires_in`, and `created => 1`; assert the stored `created` is current and `isAccessTokenExpired()` is false.

cssjoe (changes requested):
Requesting changes on head a6b4cb7.

The refresh rewrite is the right direction (fail closed on error bodies, keep credentials on transient failures, require `access_token` + `expires_in`). Two medium issues still break the renew contract, and one of them is already open from Bugbot on 23e60c7 and was not addressed in the follow-up commit.

**Please fix**

1. After a successful persist, delete `w3tcps_refresh_fail` and `w3tcps_refresh_fail_message`. Assert both are false after a later success that followed a recorded failure.
2. On successful refresh, always set `created` to `time()` so a response with a stale `created` cannot leave `isAccessTokenExpired()` true. Assert a body with `created => 1` still stores a current timestamp.
3. Add a `refresh-token-missing-refresh-token` case that clears credentials (sibling of the existing 428 not-found test).

Then ping for rereview. I did not exercise live authorization against the companion API in this pass.

PageSpeed_Api.php
```diff
@@ -261,61 +260,140 @@ public function refresh_token( $w3tc_site_id, $w3tc_pagespeed_key ) {
...
+		}
+
+		$w3tc_access_token = wp_json_encode( $response_body );
```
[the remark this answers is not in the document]
jacobd91: Addressed in d9365c82. A successful persist now deletes `w3tcps_refresh_fail` and `w3tcps_refresh_fail_message`, with a unit test covering a leftover notice from a prior failure.

PageSpeed_Api.php
jacobd91: Fixed in `9256c1f0`. Successful refresh now always sets `created` to `time()`. `test_successful_refresh_overwrites_stale_created` mocks `created => 1` and asserts a current timestamp plus `isAccessTokenExpired()` false.

PageSpeed_Api.php
jacobd91: Landed in `d9365c82` (before this rereview) and still present after merging `master` / 2.10.6. Successful persist deletes both options; `test_successful_refresh_clears_prior_failure_notice` covers fail-then-success.

tests/admin/class-w3tc-pagespeed-refresh-token-test.php
jacobd91: Added `test_missing_refresh_token_clears_credentials` in `9256c1f0`. Also adjusted the `refresh-token-google-failed` / flat Google error copy so it says authorization was kept for retry instead of telling the operator to re-authorize first.

jacobd91:
@cssjoe Ready for rereview on `9256c1f0`.

- Merged current `master` (2.10.6).
- Your item 1 (clear failure options on success) was already in `d9365c82`.
- Always overwrite `created` on successful refresh; added coverage for a stale `created => 1` body.
- Added the `refresh-token-missing-refresh-token` credential-wipe test.
- Retryable Google errors now say authorization was kept for retry instead of leading with re-authorize.

`phpunit --filter W3tc_Pagespeed_Refresh_Token_Test` — OK (8 tests, 26 assertions).

cssjoe (changes requested):
Rereviewing head `9256c1f0`. All three items from the `a6b4cb7` round are fixed and covered — verified locally, not just read.

**Verified fixed**

| Prior item | Where | Test |
|---|---|---|
| Clear both failure options on success | `PageSpeed_Api.php:325-326` | `test_successful_refresh_clears_prior_failure_notice` |
| Always stamp `created = time()` | `PageSpeed_Api.php:317` | `test_successful_refresh_overwrites_stale_created` |
| `refresh-token-missing-refresh-token` wipe | `clear_pagespeed_credentials()` | `test_missing_refresh_token_clears_credentials` |

I also checked the error-ID contract against the API companion PR: every ID that endpoint emits (`refresh-token-missing-site-id`, `-missing-w3tc-pagespeed-key`, `-not-found`, `-missing-refresh-token`, `refresh-token-google-failed`, and the nine `authorize-in/out-*` IDs) now has a mapped message here, and the retryable/irrecoverable split matches which rows that PR keeps versus abandons. `phpunit` OK (8 tests, 26 assertions), full suite OK (577 tests), `php -l` and `phpcs` clean on both changed files.

The automated passes I run locally reported nothing on this head, matching Bugbot here, so everything below came out of the manual pass.

Three new findings on this head, ordered.

**1. Medium — a permanently dead grant is retried on every General Settings render, with no cooldown**

`maybe_refresh_token()` (`PageSpeed_Api.php:230-235`) retries whenever the stored token is expired and a key is present. Keeping credentials on a retryable failure is the right call, but nothing records that the last attempt failed, so a grant Google will never renew (`invalid_grant`, or the API's 502 `refresh-token-google-failed`) is re-attempted on every construction — and `inc/options/general.php:1256` constructs `PageSpeed_Api` on every General Settings page load, with a 60-second request timeout.

This is not an edge case at rollout. Per the API PR's own deploy note, every site whose refresh token was already truncated or rotated away is in exactly this state until someone clicks Authorize, so shipping this turns their General Settings page into a blocking round trip that cannot succeed. Before this change those sites made no request at all, because the credentials had been cleared.

I confirmed the behavior rather than inferring it: three consecutive `maybe_refresh_token()` calls against a flat `invalid_grant` body produce three HTTP requests.

Fix: record a short-lived cooldown in `set_refresh_failure()` (a transient of 15 minutes or so is enough), bail early in `maybe_refresh_token()` while it is set, and delete it on success next to the two existing `delete_option()` calls. Test: after a recorded failure, two `maybe_refresh_token()` calls should produce one request.

I prototyped exactly that locally to be sure it does not fight the rest of the change — three lines, and it collapses the three doomed attempts above to one while all 8 of your tests stay green. Gating it in `maybe_refresh_token()` rather than in `refresh_token()` matters: the automatic path backs off, but an explicit re-authorization still refreshes immediately, so a user fixing the problem never waits out the cooldown.

Alternatively the API could answer a terminal `invalid_grant` with an irrecoverable ID so this side clears and prompts re-authorization — but the cooldown belongs here regardless, so this PR stays safe to merge on its own.

**2. Medium — `$w3tc_message` still has no default, which is the class the new line instantiates**

`inc/options/general.php:1267-1299` assigns `$w3tc_message` only inside the two if/elseif chains. An ID that is not in the list leaves it undefined, so PHP 8 emits `Warning: Undefined variable $w3tc_message` and `update_option()` stores `null` — the notice renders its title with a blank detail line. I reproduced both the warning and the `null` write on PHP 8.3.

That is precisely the symptom that made `authorize-out-refresh-token-missing` hard to read during the first dev authorization attempt, and mapping one more ID does not close it: the API companion PR is adding IDs on the other side of this contract right now, so the next new one lands in the same blank-message hole.

Fix: initialize `$w3tc_message` to a generic fallback (including the unrecognized ID, which is what an operator needs to report) before the chains, and assert the fallback for an unknown ID. While there, `$w3tc_authorize_error->error->id` is read without confirming `json_decode()` produced an object with that shape, so a malformed value yields "Attempt to read property on null" warnings ahead of the redirect.

**3. Medium — `reset()` keeps the three gaps this PR just closed next door**

`PageSpeed_Api.php:471-529` is the sibling consumer of the same API error shape and still has all three of the problems `refresh_token()` was rewritten to fix:

- `is_wp_error( $response )` is checked after `wp_remote_retrieve_body()`, and the transport branch returns a JSON string that `inc/options/general.php:1310-1318` discards — so a timeout during Deauthorize reports "authorization successfully reset" while the local credentials are still in place and the page still shows Deauthorize.
- HTTP status is never inspected. A 5xx or an HTML error body has no nested `error.code`, so it falls through to clearing local credentials while the remote record survives — the reverse of the bug this PR fixes, and the user can no longer revoke from W3TC.
- Unknown revoke IDs leave `$w3tc_message` undefined, same as finding 2.

Mirroring the new `set_refresh_failure()` / status-check structure here is mostly mechanical. If you would rather keep this PR scoped to renewal, that is reasonable — but please link a follow-up ticket in the description so the sibling does not sit unclaimed.

**Low — test additions worth landing with the above**

The new suite covers the paths it describes. Four gaps I probed locally, all passing against this head, so these are assertions to lock in rather than bugs:

- A non-2xx with an unrecognized (or absent) `error.id` keeps credentials and reports `Unexpected token refresh response (HTTP 500).`
- HTTP 200 carrying `access_token` but no `expires_in` is rejected without writing config at all (assert `save()` is never called, not just that the old value survives).
- A non-JSON 200 body (maintenance page) is rejected.
- `maybe_refresh_token()` makes no request when the API key is empty.

Also worth one line: `wp_json_encode()` can return `false`, and the result goes straight into `setAccessToken()`, which throws `W3TCG_Google_Auth_Exception` on a value it cannot decode. Practically unreachable, since the body already decoded successfully, but it is an uncaught-exception path on an admin page render.

I did not exercise live authorization against the companion API in this pass either.

jacobd91:
@cssjoe Ready for rereview on `bba272af`.

Addressed the three medium findings from the `9256c1f0` round:

- Automatic refresh now backs off for 15 minutes after a failure; direct `refresh_token()` remains ungated. The cooldown is cleared on success.
- Authorization errors are shape-checked and mapped through a tested fallback that includes unknown IDs.
- `reset()` now handles transport errors, non-2xx/invalid responses, and unknown IDs without clearing credentials or posting a false success.

Also added the four suggested low-cost refresh assertions (unknown 500, missing `expires_in`, non-JSON 200, empty API key).

`phpunit --filter W3tc_Pagespeed_Refresh_Token_Test` — OK (17 tests, 51 assertions). `php -l` and `phpcs` clean on all three changed files.

cssjoe (changes requested):
Rereviewing head `bba272af`. All three medium findings from the `9256c1f0` round are genuinely closed — I verified each one by running it, not by reading the diff.

**Verified fixed**

| Prior item | Where | Evidence |
|---|---|---|
| No cooldown on a dead grant | `PageSpeed_Api.php:240-248`, `353-357` | Cooldown measures 900s; two `maybe_refresh_token()` calls inside the window make one request, a third after it lapses makes a second; `refresh_token()` success deletes `w3tcps_refresh_retry_after` |
| `$w3tc_message` had no default | `get_authorize_failure_message()` | Unknown ID returns the ID in the text; `'not-json'` returns the `unknown` fallback; `isset( $error->error->id )` covers the shape |
| `reset()` sibling gaps | `PageSpeed_Api.php:546-601` | `is_wp_error()` now precedes body handling, status is inspected, returns `bool`, and `general.php:1289` only posts success on `true` |

The error-ID contract is also complete now. Every ID the companion endpoint emits — six `refresh-token-*`, four `revoke-token-*`, nine `authorize-*` — has a mapped message here, plus a fallback. `phpunit --filter W3tc_Pagespeed_Refresh_Token_Test` OK (17 tests, 51 assertions), full suite OK (586 tests, 2317 assertions, 2 pre-existing skips), `php -l` and `phpcs` clean on all three files.

Not merge-ready — one high and two medium findings on this head, all in the newly added failure classification. The refresh half is solid; the revoke half is built against a contract the endpoint does not implement.

**Which pass produced which finding**

| Pass | Result |
|---|---|
| Local automated review | No findings on this head — matches Bugbot here. Does not offset the three below. |
| Automated review of changed surfaces | No blocking findings. Does not offset the three below. |
| Manual product pass | All three findings below |

**Still open from prior rounds (carried forward):** the `wp_json_encode()` return value at `PageSpeed_Api.php:334`, listed under Low.

**1. High — `reset()` reads the revoke outcome from the redirect target's status code, not from the revoke result**

`revokeToken()` in the companion service returns `redirect(...)` on **every** path — all four error branches and the success branch. Only `refreshToken()` uses `response( $body, $status )`. So the JSON-status contract that `PageSpeed_Api.php:575-595` was just written against does not exist on this endpoint, and `wp_remote_get()` is called with the default `redirection => 5`, so the plugin follows the 302 and classifies the revoke on whatever page the chain ends on — the site's own `wp-login.php`, since the API redirects back to an authenticated admin URL.

I ran both directions:

- A revoke **error** carried in the 302 query string, chain ending on a login page (HTTP 200 HTML): `reset()` returns `true`, credentials are cleared, and General Settings posts "authorization successfully reset" — while the remote record is still active. This is the same false-success the PR removes on the refresh side.
- A **successful** revoke whose redirect target answers 503: `reset()` returns `false`, credentials are kept, and the notice reads `Unexpected token revocation response (HTTP 503, error unknown).`

The second one is a new regression, and it strands the operator. `general.php:1302` renders Deauthorize only while `! isAccessTokenExpired()`, so during the token's valid hour a failing Deauthorize is the *only* control on the page — there is no Authorize button to fall back to, and each retry fails the same way. Before this PR `reset()` always cleared local config, so Deauthorize always worked locally regardless of the remote answer.

All four `get_revoke_failure_message()` IDs are unreachable as written: the ID only ever arrives in the `w3tc_authorize_error` query parameter of the redirect, and `PageSpeed_Api.php` never reads that parameter.

Fix, in preference order:

1. Have the companion PR answer `revokeToken()` with `response( $body, $status )` the way `refreshToken()` already does, and land the two together. That makes the existing plugin code correct as-is and is the smaller change. Please mark the dependency as blocking on both PRs — as it stands this PR's revoke handling cannot behave correctly against the deployed endpoint.
2. Failing that, pass `'redirection' => 0` and read the error ID out of the `Location` header's `w3tc_authorize_error` parameter.

Either way, treat `revoke-token-not-found` and the three `*-missing` IDs as irrecoverable and clear local credentials — there is nothing left to revoke remotely, and keeping them is what produces the dead-end. Test: assert `reset()` clears credentials for `revoke-token-not-found`, and that a 5xx from the redirect target does not leave the operator with credentials and no Authorize button.

While you are in that endpoint: its four revoke error branches use `$returnUrl` before it is assigned (only set at line 252, after all four), so each one emits an undefined-variable warning and redirects to a relative URL. That is the companion PR's to fix, not this one, but it is why the error path resolves the way it does.

**2. Medium — the stale-notice class is closed only for the refresh-success path**

`refresh_token()` success deletes `w3tcps_refresh_fail`, `w3tcps_refresh_fail_message`, and `w3tcps_refresh_retry_after`. The two sibling success paths do not:

- `general.php:1281-1287` (authorize returns with a new token) saves config and redirects without touching them.
- `reset()` success (`PageSpeed_Api.php:597-600`) clears credentials without touching them.

Because `errors_notice()` is a `switch ( true )` that checks `w3tcps_authorize_success` before `w3tcps_refresh_fail`, the success notice displays and deletes itself while the older failure survives to the next page load. I rendered it twice against seeded options: first render is "authorization successfully reset", second is "Google PageSpeed access token refresh failed." — after the operator has already fixed the problem. That is exactly the finding Bugbot filed on `23e60c7c`, in the paths the fix did not reach.

The refresh options are also set during page-content render, which runs after `admin_notices` has fired, so a failure recorded while rendering General Settings is always displayed on the *following* load — which is usually the load right after the operator re-authorizes.

Fix: delete the three refresh options (and the two revoke ones) in the authorize-success branch and on successful `reset()`; a small private `clear_failure_notices()` on `PageSpeed_Api` would keep the three call sites honest. Test: seed `w3tcps_refresh_fail`, run a successful `reset()`, assert it is gone.

**3. Medium — "token present but expired" is now durable, and the two front-end consumers do not recognize it**

Keeping credentials on a retryable failure is the right call, but it makes a state permanent that previously could not persist: `widget.pagespeed.access_token` non-empty and expired. `general.php:1302` handles it — it checks `isAccessTokenExpired()` and renders Authorize. The other two consumers only check for emptiness:

- `PageSpeed_Widget.php:95` and `:141`
- `PageSpeed_Page.php:102`

Both fall through to `analyze()` with a known-expired token, so the dashboard widget and the PageSpeed tool report a raw "API request error!" with the upstream response code instead of the authorize prompt they used to show when credentials were wiped. During a 15-minute cooldown that is the only thing the operator sees on those two screens, and it points at Google rather than at the Authorize button one page over.

Fix: gate those three call sites on `isAccessTokenExpired()` as well, so they emit the existing `missing_token` copy. Test: expired token in config, assert the widget AJAX payload carries `missing_token` rather than an `error`.

**Low — carried forward and small**

- `get_authorize_failure_message()` assumes a string. `Util_Request::get()` returns the raw request value, which can be an array, so `json_decode()` gets a non-string and throws `TypeError: json_decode(): Argument #1 ($json) must be of type string, array given`. Pre-existing — the old inline code had the same shape — but the new helper is the natural place for a one-line `is_string()` guard, and the two fallback branches would then have a single well-typed input to work from. Please also constrain the ID the two fallbacks interpolate to `[a-z0-9-]+` (or drop it) while you are in there; see the ticket for the reasoning.
- `wp_json_encode()` can return `false` at `PageSpeed_Api.php:334`, and the result goes straight into `setAccessToken()`, which throws on a value it cannot decode. Still open from the last round; practically unreachable since the body already decoded, but it is an uncaught throw on an admin render.
- `clear_pagespeed_credentials()` leaves `w3tcps_refresh_retry_after` set. Harmless today because the key is empty afterwards, but it survives re-authorization.
- Three test gaps worth closing with the above: `revoke-token-not-found`; a cooldown that lapses and then succeeds (only the "no second request" half is asserted); and a stale notice cleared by a successful revoke.

I did not exercise live authorization against the companion service in this pass.

jacobd91:
@cssjoe Follow-up feedback is addressed in `f93cd385`:

- Expired tokens now follow the same authorization-required path as missing tokens in the dashboard widget, widget AJAX endpoint, and PageSpeed page—even when cached data exists.
- Successful authorization, refresh, and revoke clear stale refresh/revoke notices and retry cooldown state; irrecoverable credential cleanup also clears the cooldown.
- Revoke now distinguishes terminal remote-state errors (clear local credentials) from transient/API failures (preserve credentials for retry).
- Authorization fallback IDs are allowlisted to safe characters before interpolation, and token JSON encoding is guarded.
- Added regression coverage for cooldown cleanup, stale notices, terminal/transient revoke responses, malicious IDs, and the expired-client state.

Validation: PHPCS passed; `W3tc_Pagespeed_Refresh_Token_Test` passed (18 tests, 67 assertions).

jacobd91:
@cssjoe Small follow-up in `dbc79820`, paired with `4fcfb3a` on the api half.

The API now guards the `authorizeOut()` Google call and redirects with a new `authorize-out-google-failed` id instead of returning a Lumen 500, so a Google 4xx during Authorize — an expired or replayed `code`, or consent revoked mid-flow — comes back as a parseable error. `get_authorize_failure_message()` maps that id to plain text rather than falling through to the generic unknown-id fallback, with a test asserting the raw id never reaches the notice.

Also worth noting for the revoke contract we landed last round: the API no longer returns 502 when Google refuses an already-expired access token. It deactivates the grant and answers 200, so `revoke-token-google-failed` on this side now only means a genuine transport failure, which is the transient case the plugin already keeps credentials for.

Validation: `phpunit --filter W3tc_Pagespeed_Refresh_Token_Test` → OK (19 tests, 69 assertions); `php -l` and `phpcs` clean.

PageSpeed_Api.php
```diff
@@ -412,41 +596,110 @@ public function reset() {
...
+			isset( $response_body['error'] )
+		) {
+			if (
+				in_array(
+					$w3tc_error_id,
+					array(
+						'revoke-token-access-token-missing',
+						'revoke-token-site-id-missing',
+						'revoke-token-api-key-missing',
+						'revoke-token-not-found',
```
cssjoe, line 629: **Medium — three of these four IDs are not a completed deauthorization.**

`revoke-token-access-token-missing`, `revoke-token-site-id-missing`, and `revoke-token-api-key-missing` all come from api2's parameter guards, before it ever loads the `google_auth` row (companion PR 18, `revokeToken()` lines 195-231 — three JSON 400s). The grant is untouched. Taking this branch wipes the local token and key, returns `true`, and `inc/options/general.php:1290` then posts "authorization successfully reset" — so the operator is told the revoke landed, has no credentials left to retry with, and the remote record stays active. That is the same false success this PR removes on the refresh side.

Only `revoke-token-not-found` earns the early `true`: api2 returns it after the lookup misses, so there is genuinely nothing left to revoke.

It also makes the copy unreachable — `get_revoke_failure_message()` maps all three `*-missing` IDs (`PageSpeed_Api.php:679-686`) and nothing can ever reach those cases.

Fix: narrow the check to `'revoke-token-not-found' === $w3tc_error_id` and let the three `*-missing` IDs fall through to `set_revoke_failure()` / `return false`.

Test: mock HTTP 400 with `revoke-token-access-token-missing`; assert `reset()` is `false`, the stored token and key are unchanged, and `w3tcps_revoke_fail_message` is `No access token provided for revoke!`. A sibling case for one of the other two `*-missing` IDs keeps the class closed, the way `test_missing_refresh_token_clears_credentials` does for the refresh pair.

PageSpeed_Widget.php
```diff
@@ -127,6 +128,25 @@ public function w3tc_ajax_pagespeed_widgetdata() {
...
+						__(
+							'Before you can get started using the Google PageSpeed tool, you’ll first need to authorize access. Please click %1$s.',
+							'w3-total-cache'
+						),
+						'<a href="' . esc_url( Util_Ui::admin_url( 'admin.php?page=w3tc_general#google_pagespeed' ) ) . '" target="_blank">' . esc_html__( 'here', 'w3-total-cache' ) . '</a>'
+					),
+				)
+			);
+			return;
+		}
```
cssjoe, line 149: **Medium — the expiry gate runs ahead of the cache, so a cooling-down site loses results it already has.**

Three call sites now construct `PageSpeed_Api` and check `isAccessTokenExpired()` before the `w3tc_pagespeed_data_*` lookup:

- here (`PageSpeed_Widget.php:133-149`)
- `PageSpeed_Page.php:91-109`
- `PageSpeed_Widget.php:95-97`, on the widget HTML itself

Gating `analyze()` on expiry is right — that was my finding last round and it is fixed. The problem is where the gate sits. Because this PR deliberately keeps credentials through a retryable failure, "token stored but expired" is now a durable state that lasts at least the 15-minute cooldown. During that window all three screens return `missing_token` even when a valid cached score from minutes earlier is sitting in options, and the operator is told to authorize something that is already authorized. Before this change the token was either wiped (so the prompt was correct) or refreshed on the spot.

`PageSpeed_Widget.php:95-97` is the worst of the three: constructing `PageSpeed_Api` there runs `maybe_refresh_token()` on every dashboard render, so outside the cooldown the dashboard blocks on a 60-second refresh request, and inside it the whole widget is replaced by the authorize prompt rather than showing the last known scores.

Fix: move the cache lookup above the construction in both AJAX handlers and keep the expiry check on the cache-miss path immediately before `analyze()`. For the widget HTML, gate on an empty token string as it did before — the AJAX handler behind it is what needs to distinguish expired from missing.

Test: seed a fresh `w3tc_pagespeed_data_{home_url}` option, put an expired token in config, and assert the widget AJAX payload carries the cached data rather than `missing_token`; and with no cache present, assert it still carries `missing_token` and makes no `analyze()` call.

PageSpeed_Api.php
```diff
@@ -412,41 +596,110 @@ public function reset() {
...
 
-		$this->w3tc_config->set( 'widget.pagespeed.access_token', '' );
-		$this->w3tc_config->set( 'widget.pagespeed.w3tc_pagespeed_key', '' );
-		$this->w3tc_config->save();
+		return sprintf(
+			/* translators: 1: HTTP response code, 2: API error ID. */
+			__( 'Unexpected token revocation response (HTTP %1$d, error %2$s).', 'w3-total-cache' ),
+			$response_code,
+			'' !== $w3tc_error_id ? $w3tc_error_id : __( 'unknown', 'w3-total-cache' )
+		);
```
cssjoe, line 703: **Low — this fallback still interpolates the API error ID as it arrives.**

`get_authorize_failure_message()` now constrains its ID to `[a-z0-9-]+` before interpolating (`PageSpeed_Api.php:506-508`) and substitutes `unknown` otherwise. This sibling fallback does not, so the two halves of the same contract disagree on what an unrecognized ID may contain. Reasoning is on the ticket.

Fix: apply the same `preg_match( '/\A[a-z0-9-]+\z/', $w3tc_error_id )` check here and fall back to the existing `unknown` string, or wrap the value in `esc_html()`.

Test: an unrecognized ID carrying markup returns the `unknown` placeholder, and the rendered notice contains no tags — the assertion `test_authorize_failure_message_has_safe_fallback` already makes on the authorize side.

cssjoe (changes requested):
Rereviewing head `dbc79820`.

Not merge-ready — two medium findings on this head, both of them the Bugbot threads opened on `f93cd38` that this commit did not close. Everything below I confirmed by running it against this head, not by reading the diff.

**Verified fixed since `bba272af`**

| Prior item | Where | Evidence |
|---|---|---|
| High — revoke outcome read from the redirect target | api2 PR 18 @ `4fcfb3a` | `revokeToken()` now answers `response()->json( $body, $status )` on every branch, so the JSON-status contract this side was written against exists. Land the two PRs together. |
| Stale notices on the sibling success paths | `PageSpeed_Api.php:562-568`, `645-646`, `inc/options/general.php:1285` | `clear_failure_notices()` runs on refresh persist, successful `reset()`, and authorize return; `test_successful_revoke_clears_credentials` seeds all five options and asserts each is gone |
| `wp_json_encode()` can return `false` | `PageSpeed_Api.php:335-340` | Encode failure records a refresh failure instead of reaching `setAccessToken()` |
| Cooldown surviving a credential wipe | `PageSpeed_Api.php:418` | `clear_pagespeed_credentials()` deletes `w3tcps_refresh_retry_after`; both wipe tests assert it |
| `get_authorize_failure_message()` input typing | `PageSpeed_Api.php:500-508` | `is_string()` guard plus the `[a-z0-9-]+` constraint on the interpolated ID, covered by `test_authorize_failure_message_has_safe_fallback` |

`phpunit --filter W3tc_Pagespeed_Refresh_Token_Test` OK (19 tests, 69 assertions), `php -l` and `phpcs` clean on all three changed files. Travis, phpcs, and semgrep are green on this head.

**Which pass produced which finding**

| Pass | Result |
|---|---|
| Local automated review | One medium: the expiry gate ahead of the cache lookup. Does not offset the revoke finding. |
| Automated review of changed surfaces | No medium-or-higher findings. Does not offset the two below. |
| Bugbot here | Both mediums below, filed on `f93cd38` and still open. The check on this head is Neutral, so it did not re-file them. |
| Manual product pass | Confirms both, plus the low. |

**Please fix, in order**

1. Medium — `PageSpeed_Api.php:622-636`: treat only `revoke-token-not-found` as a completed deauthorization. The three `*-missing` IDs mean the request never reached the grant.
2. Medium — `PageSpeed_Widget.php:133-149`, `:95-97`, and `PageSpeed_Page.php:91-109`: read the cache before constructing `PageSpeed_Api` or checking expiry, and stop replacing the whole dashboard widget with the authorize prompt while a token is stored.
3. Low — `PageSpeed_Api.php:698-703`: constrain the interpolated revoke error ID the same way the authorize fallback now does.

Detail and the tests I would want with each are inline.

**Still open from prior rounds (carried forward)**

One low from the `bba272af` round remains beyond item 3: the cooldown test asserts only the "no second request inside the window" half. A lapsed cooldown that then succeeds is still uncovered, and that is the half that proves an operator is not stuck for 15 minutes longer than intended.

I did not exercise live authorization against the companion service in this pass either.

PageSpeed_Api.php
jacobd91: Fixed in `cd839a3a`. Only `revoke-token-not-found` now clears credentials and returns true. The three `*-missing` IDs fall through to `set_revoke_failure()` / `false`, with tests for access-token-missing and api-key-missing.

PageSpeed_Widget.php
jacobd91: Fixed in `cd839a3a`. Both AJAX handlers read the cache before constructing `PageSpeed_Api`. The dashboard widget HTML gates only on an empty token string. Added coverage that an expired token still returns cached widget data, and without cache still returns `missing_token` with no outbound request.

PageSpeed_Api.php
jacobd91: Fixed in `cd839a3a`. The revoke fallback now applies the same `[a-z0-9-]+` constraint as authorize. Markup in an unrecognized ID becomes `unknown` and never reaches the notice.

PageSpeed_Api.php
```diff
@@ -412,41 +593,110 @@ public function reset() {
...
-				$w3tc_message = __( 'No access token provided for revoke!', 'w3-total-cache' );
-			} elseif ( 'revoke-token-api-key-missing' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No W3TC API key provided for revoke!', 'w3-total-cache' );
-			} elseif ( 'revoke-token-not-found' === $response_body['error']['id'] ) {
-				$w3tc_message = __( 'No matching Google access record found for W3TC API key!', 'w3-total-cache' );
+			) {
+				$this->clear_pagespeed_credentials();
+				self::clear_failure_notices();
+				return true;
 			}
```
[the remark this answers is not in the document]
jacobd91: Addressed in `cd839a3a` with the same change as the later review: only `revoke-token-not-found` is treated as a completed deauthorization.

PageSpeed_Widget.php
```diff
@@ -92,8 +92,9 @@ public function wp_dashboard_setup() {
 	public function widget_pagespeed() {
 		$w3tc_config       = Dispatcher::config();
 		$w3tc_access_token = $w3tc_config->get_string( 'widget.pagespeed.access_token' );
+		$pagespeed_api     = empty( $w3tc_access_token ) ? null : new PageSpeed_Api( $w3tc_access_token );
 
-		if ( empty( $w3tc_access_token ) ) {
+		if ( empty( $pagespeed_api ) || $pagespeed_api->client->isAccessTokenExpired() ) {
```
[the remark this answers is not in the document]
jacobd91: Addressed in `cd839a3a`. Cache is consulted before expiry, and the widget shell no longer constructs `PageSpeed_Api` on dashboard render.

jacobd91:
Follow-up from the `dbc79820` rereview is in `cd839a3a`.

- Only `revoke-token-not-found` is treated as a completed deauthorization. The three `*-missing` IDs keep credentials and surface the mapped failure notice.
- Widget and page AJAX read the cache before constructing `PageSpeed_Api` or checking expiry. The dashboard widget HTML gates only on an empty token string.
- Revoke fallback IDs use the same `[a-z0-9-]+` constraint as authorize.
- Also covered the carried-forward cooldown case: a lapsed retry window allows a successful refresh.

Validation: `phpunit --filter W3tc_Pagespeed_Refresh_Token_Test|W3tc_Pagespeed_Widget_Cache_Test` — OK (26 tests, 96 assertions); phpcs clean.

PageSpeed_Api.php
```diff
@@ -261,59 +273,149 @@ public function refresh_token( $w3tc_site_id, $w3tc_pagespeed_key ) {
...
+	 *
+	 * @since X.X.X
+	 *
+	 * @return void
+	 */
+	private function clear_pagespeed_credentials() {
+		$this->w3tc_config->set( 'widget.pagespeed.access_token', '' );
+		$this->w3tc_config->set( 'widget.pagespeed.w3tc_pagespeed_key', '' );
+		$this->w3tc_config->save();
+		delete_option( 'w3tcps_refresh_retry_after' );
```
jacobd91, line 404: **Low — leftover of serving cache before the expiry gate.**

Deauthorize / irrecoverable refresh now clears the access token and API key here, but not `w3tc_pagespeed_data_*`. Both AJAX handlers read that cache before constructing `PageSpeed_Api`, so the PageSpeed page can keep showing scores after disconnect. The dashboard widget HTML already treats an empty token as authorization-required, so the two surfaces can disagree for the cache lifetime.

GitHub Bugbot filed this on the readers; close the class on the writer: delete the known PageSpeed result options (home URL and any URL-keyed entries you already persist) from `clear_pagespeed_credentials()` / successful `reset()`.

Test: seed a fresh cache option, run a successful `reset()`, assert the cache option is gone and widget/page AJAX returns `missing_token` with no outbound request.

PageSpeed_Page.php
```diff
@@ -117,8 +118,7 @@ public function w3tc_ajax_pagespeed_data() {
 				return;
```
jacobd91: **Medium — cache-miss still labels a kept-but-expired token as missing authorization.**

Gating `analyze()` on expiry after `PageSpeed_Api` construction is right (constructor already ran `maybe_refresh_token()`). The copy is not: this payload is the same "authorize access" string used when no token exists.

Retryable refresh failures keep credentials and start a 15-minute cooldown. During that window, with no cache, the PageSpeed page and the widget AJAX handler (`PageSpeed_Widget.php` sibling of this block) tell the operator to authorize something that is still stored.

Fix: if the config token string is non-empty but the client is still expired after construction, return a distinct retry/refresh-failed payload (and keep `missing_token` for a truly empty token). Cover both AJAX handlers.

Test: expired stored token, no cache, cooldown in the future; assert the payload is not `missing_token` and that no `analyze()` / Google request is made.

inc/options/general.php
```diff
@@ -1302,15 +1282,17 @@ class="w3tc-ignore-change" type="text"
 			$this->_config->set( 'widget.pagespeed.access_token', $w3tc_access_token );
 			$this->_config->set( 'widget.pagespeed.w3tc_pagespeed_key', $w3tc_pagespeed_key );
 			$this->_config->save();
+			PageSpeed_Api::clear_failure_notices();
```
jacobd91: **Medium — leftover of always stamping `created` on persist.**

Successful `refresh_token()` always sets `$response_body['created'] = time()` because `W3TCG_Google_Auth_OAuth2::isAccessTokenExpired()` treats a missing `created` as expired. This authorize return still writes `w3tc_access_token` verbatim (the three `set`/`save` lines above), then only clears failure notices.

If the W3 API omits `created`, a brand-new grant looks expired: constructor refresh may fire immediately, and a cache-miss AJAX path can return `missing_token` while credentials are stored.

Fix: decode the token JSON, require a string `access_token` and numeric `expires_in`, set `created` to `time()`, re-encode, then save (same contract as refresh, including a `wp_json_encode()` failure path). `clear_failure_notices()` can stay.

Test: persist a grant whose JSON has `access_token` / `expires_in` but no `created`; assert stored `created` is current and `isAccessTokenExpired()` is false without a refresh request.

jacobd91 (commented):
Cannot request changes on my own PR, so this is a review comment with the leftover-class items from the last round. PHPUnit for the PageSpeed refresh/widget tests is green (26 tests, 96 assertions). Admin page smoke for dashboard / general / PageSpeed also passed.

**Please address before merge**
1. Authorize persist still does not stamp `created` (refresh already does).
2. Cache-miss AJAX still uses the authorize-access `missing_token` copy when a stored token is expired after refresh/cooldown.

**Non-blocking**
3. `clear_pagespeed_credentials()` / successful `reset()` should drop `w3tc_pagespeed_data_*` so cache-first AJAX cannot serve scores after disconnect (same class as the open Bugbot thread on the readers).

Please close the class in the writers (authorize persist + credential clear) and split expired-vs-missing operator copy on both AJAX handlers.

inc/options/general.php
jacobd91: Fixed in 12cd8802. Authorize persist now runs through `PageSpeed_Api::prepare_access_token_json()`, the same contract refresh uses: it decodes the payload, requires a string `access_token` and numeric `expires_in`, stamps `created` with `time()`, and re-encodes, returning `false` (including on `wp_json_encode()` failure) instead of saving. On `false` the handler writes the authorize-failure notices and skips the three `set`/`save` lines; `clear_failure_notices()` still runs on success. `refresh_token()` now calls the same helper, so there is one persist contract.

Test: `test_prepare_access_token_json_stamps_created()` persists a grant with `access_token`/`expires_in` and no `created`, asserts the stored `created` is current and `isAccessTokenExpired()` is false with no outbound request. `test_prepare_access_token_json_rejects_unusable_payloads()` covers non-JSON, missing `expires_in`, and a non-string `access_token`.

PageSpeed_Page.php
jacobd91: Fixed in 12cd8802. Both AJAX handlers now split the two states. An empty stored token returns `missing_token` with the authorize-access copy; a token that is still expired after construction (retryable failure, cooldown active) returns a new `refresh_failed` key whose copy says access is authorized but could not be renewed, W3TC will retry, and offers the reauthorize link. Both strings moved to `PageSpeed_Api::get_authorize_required_message()` / `get_refresh_pending_message()` so widget and page cannot drift. Each handler also re-reads the config token after construction, since an irrecoverable refresh clears credentials mid-request; that case falls to `missing_token`. `PageSpeed_Widget_View.js` and `PageSpeed_Page_View.js` render the new key in the existing notice container.

Test: `test_widget_ajax_refresh_failed_when_expired_without_cache()` and `test_page_ajax_refresh_failed_when_expired_without_cache()` assert the payload is not `missing_token` and that no request is made; `test_widget_ajax_missing_token_without_stored_token()` and the page equivalent keep `missing_token` for an empty token.

PageSpeed_Api.php
jacobd91: Fixed in 12cd8802. `clear_pagespeed_credentials()` now calls `PageSpeed_Api::clear_pagespeed_cache()`, which deletes the home-URL keyed option plus every `w3tc_pagespeed_data_%` option (selected via `$wpdb->esc_like()` and removed with `delete_option()` so the options cache stays coherent). Both `reset()` success paths and the irrecoverable refresh path already route through `clear_pagespeed_credentials()`, so the writer side is closed and cache-first AJAX cannot serve scores after disconnect.

Test: `test_successful_revoke_clears_cached_results()` seeds both a home-URL and a URL-keyed cache option, runs a successful `reset()`, and asserts both are gone. With the cache cleared, the AJAX paths fall through to the token check, covered by the widget/page tests above.

PageSpeed_Page.php
```diff
@@ -100,8 +100,9 @@ public function w3tc_ajax_pagespeed_data() {
 		if ( is_null( $api_response ) ) {
 			$w3tc_config       = Dispatcher::config();
 			$w3tc_access_token = ! empty( $w3tc_config->get_string( 'widget.pagespeed.access_token' ) ) ? $w3tc_config->get_string( 'widget.pagespeed.access_token' ) : null;
+			$w3tc_w3_pagespeed = empty( $w3tc_access_token ) ? null : new PageSpeed_Api( $w3tc_access_token );
 
-			if ( empty( $w3tc_access_token ) ) {
+			if ( empty( $w3tc_w3_pagespeed ) || $w3tc_w3_pagespeed->client->isAccessTokenExpired() ) {
```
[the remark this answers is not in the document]
jacobd91: Fixed in 12cd8802 on the writer side. Deauthorize and irrecoverable refresh now delete every `w3tc_pagespeed_data_*` option via `PageSpeed_Api::clear_pagespeed_cache()`, so there is no cache left to serve once the access token is gone. The readers additionally re-read the stored token after constructing `PageSpeed_Api` and return `missing_token` when it is empty, so the page and the dashboard widget agree. Covered by `test_successful_revoke_clears_cached_results()`.

jacobd91:
All four open threads are addressed in 12cd8802 and resolved.

**Blocking items**
1. **Authorize persist stamps `created`.** New `PageSpeed_Api::prepare_access_token_json()` is the single persist contract for both authorize and refresh: validate string `access_token` + numeric `expires_in`, stamp `created`, re-encode, and return `false` on any problem (including `wp_json_encode()` failure). Authorize writes the failure notices and skips the save when it returns `false`.
2. **Expired vs unauthorized copy.** Both AJAX handlers now return `missing_token` only for an empty token, and a new `refresh_failed` key with retry copy when a stored token is still expired after construction. Copy lives in `PageSpeed_Api::get_authorize_required_message()` / `get_refresh_pending_message()`, and both view scripts render the new key.

**Non-blocking / open Bugbot thread**
3. **Cache cleared with credentials.** `clear_pagespeed_credentials()` now calls `PageSpeed_Api::clear_pagespeed_cache()`, deleting the home-URL option and every `w3tc_pagespeed_data_%` option. Both `reset()` success paths and the irrecoverable refresh path already route through it, so cache-first AJAX cannot serve scores after a disconnect. Handlers also re-read the stored token after construction, so a mid-request credential clear reports `missing_token` rather than serving stale state.

**Validation:** `phpunit --filter W3tc_Pagespeed_Refresh_Token_Test|W3tc_Pagespeed_Widget_Cache_Test` is green — 32 tests, 123 assertions (was 26/96). New coverage: `created` stamping and rejection of unusable payloads, widget and page `refresh_failed` vs `missing_token`, and cache clearing on successful revoke. PHPCS and `php -l` clean on all changed PHP.

cssjoe (approved):
Approving head `12cd8802`.

Prior review items on this PR are closed and covered: fail-closed refresh, `created` stamped on persist (authorize and refresh share `prepare_access_token_json()`), cooldown on automatic retry, credential wipe only for irrecoverable refresh IDs, `reset()` HTTP-aware with only `revoke-token-not-found` treated as completed deauthorization, stale notices cleared on success paths, expired vs missing copy on cache-miss AJAX, and cached results dropped when credentials are cleared.

Verified locally: `phpunit --filter W3tc_Pagespeed_Refresh_Token_Test|W3tc_Pagespeed_Widget_Cache_Test` OK (32 tests, 123 assertions); `php -l` and phpcs clean on the changed PHP; admin smoke for dashboard, General Settings, and the PageSpeed page passed. Travis, phpcs, and semgrep are green on this head.

Non-blocking leftover: cache-first AJAX still does not require a stored access token before returning a warm `w3tc_pagespeed_data_*` entry. Writers now delete those options on disconnect, so the deauthorize path is closed. A leftover cache plus an empty token can still return scores on the widget AJAX path. Optional follow-up: gate the cache read on a non-empty token and assert empty token + matching warm cache returns `missing_token`.

Land with the companion API so revoke answers JSON status rather than a redirect. Live authorize/refresh against that service remains the acceptance item from the test plan.

PR #1388, opened 2026-09-04
ENG3-884: Omit development-only files from the public release ZIP

cssjoe:
https://imh-internal.atlassian.net/browse/ENG3-884

## Summary
- Strip development-only paths from the Travis/wordpress-tag-sync release tree before the public ZIP is built.
- Share that list in `bin/release-strip-dev.sh` and fail if any listed path remains.
- Apply the same strip and ZIP checks in `bin/build-release-zip.sh`. The list only names paths that exist in this repo (plus Travis-generated `codecov` / `coverage.xml`).

## Test plan
- [ ] Confirm `bin/release-strip-dev.sh` removes `ci/`, `policies/`, `rules/`, `tmp/`, `phpcs-rules/`, `qa/`, `tests/`, and the listed tooling files, and keeps `inc/`, `lib/`, `pub/`, `ini/`, `languages/`, `wp-content/`, `vendor/`, and `extension-example/`.
- [ ] Run `yarn run build:release <version>` and confirm the ZIP does not contain those development paths.
- [ ] On the next tagged Travis build, confirm the GitHub Releases ZIP matches the same production tree.

### DIFF
--- a/bin/build-release-zip.sh
+++ b/bin/build-release-zip.sh
@@ -234,7 +234,7 @@ if [[ -n "$forbidden" ]]; then
 fi
 
 forbidden_files="$(
-	unzip -l "$OUTPUT" | awk '{print $4}' | grep -E "^${PLUGIN_SLUG}/(\.sec-project\.yaml|phpcs\.xml|phpunit\.xml|AGENTS\.md|CLAUDE\.md)$" || true
+	unzip -l "$OUTPUT" | awk '{print $4}' | grep -E "^${PLUGIN_SLUG}/(\.jshintrc|\.sec-project\.yaml|AGENTS\.md|CLAUDE\.md|codecov|coverage\.xml|phpcs\.xml|phpunit\.xml)$" || true
 )"
 if [[ -n "$forbidden_files" ]]; then
 	echo 'error: ZIP contains development-only files' >&2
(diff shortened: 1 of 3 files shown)


jacobd91 (approved):
Reviewed the packaging changes on this branch.

The shared strip list plus fail-closed leftover check, extra rsync excludes, and ZIP forbidden-path checks cover the development trees named in the ticket (`ci`, `policies`, `rules`, `tmp`, `phpcs-rules`, `qa`, `tests`) while leaving runtime trees (`inc`, `lib`, `pub`, `ini`, `languages`, `wp-content`, production `vendor`, `extension-example`) in the artifact. Local `bin/build-release-zip.sh` produced a ZIP with none of those top-level development paths.

Non-blocking follow-ups (same packaging class, not merge blockers for this ticket):
- `--exclude='vendor/'` does not skip a `vendor` *symlink*, so staging can inherit the link and `composer install --no-dev` can mutate the link target. Prefer excluding `vendor` as well as `vendor/`.
- Gitignored local files such as `.phpunit.result.cache` are not in the strip/ZIP denylist and can land in a locally built ZIP. Travis checkouts usually will not have that file.
- The ZIP denylist is only in `bin/build-release-zip.sh`. The Travis `bin/release.sh` path still relies on strip + wordpress-tag-sync without an equivalent post-ZIP assertion.

Acceptance of tagged Travis GitHub Release ZIP remains for the next tag, as in the test plan.

PR #1395, opened 2026-09-15
ENG3-1052: Remove expired role cookie compatibility path

jacobd91:
https://imh-internal.atlassian.net/browse/ENG3-1052

## Summary
- Remove the expired page-cache role-cookie compatibility path that shipped for the 2.10.0 window.
- Keep current role-cookie names for rewrite reject lists and early cache-auth reads.
- Skip page-cache writes when the loaded WordPress user has a configured rejected role, so leftover cookies cannot store a privileged response.

## Test plan
- [ ] PHPUnit `W3tc_Util_Cookie` tests pass.
- [ ] With “Don’t cache pages for following user roles” enabled for administrator, an administrator session still sees uncached/personalized output after login.
- [ ] A subscriber (or logged-out visitor) still receives a cached public page.
- [ ] After deploy, regenerate page-cache rewrite rules so disk-enhanced installs match PHP.
- [ ] AWS Puppeteer `pagecache/user-roles` spec (release matrix).


cssjoe (approved):
Approved on `aacc329`.

Local review (Bugbot + product + security) found nothing that should block merge. PHPUnit `W3tc_Util_Cookie` is green (11 tests). On a local disk-enhanced install with “Don’t cache pages for following user roles” set for administrator, an administrator request stays uncached (`Rejected user role is logged in`); a subscriber and a logged-out request still receive the cached public page. Site settings were restored afterward.

GitHub Bugbot’s session-log / `tmp/feature-docs` finding is not in this diff (five production/test files only). The remaining changelog note about the old grace switch belongs on the next release branch, not this PR.

After merge/deploy: regenerate page-cache rewrite rules so disk-enhanced reject lists match PHP. AWS Puppeteer `pagecache/user-roles` remains the release-matrix check.

PR #1396, opened 2026-09-15
ENG3-1065 Persist WP-Cron purge schedule settings

jacobd91:
https://imh-internal.atlassian.net/browse/ENG3-1065

## Summary
- Persist Purge via WP-Cron Enable, Start Time, and Interval on General and the per-cache settings pages.
- Cover all five schedule groups (all caches, page, database, minify, object cache) with schema registration and PHPUnit save-path tests.

## Test plan
- [ ] On Performance → General Settings, enable Purge via WP-Cron, set Start Time to 3:00 AM, Interval to Daily, save. After reload the checkbox stays enabled and the time/interval keep those values.
- [ ] Repeat with Twice Daily, save again, and confirm the interval still shows Twice Daily.
- [ ] Disable the WP-Cron event, save, and confirm it stays disabled.
- [ ] Spot-check the same Enable / Start Time / Interval controls on Page Cache, Database Cache, Minify, and Object Cache advanced settings.
- [ ] PHPUnit: `W3tc_Wp_Cron_Config_Test`

Made with [Cursor](https://cursor.com)


jacobd91:
Addressed in 1e1c9dc6.

Partly accurate. The named paths were never part of this PR's diff — it contains only `ConfigKeys.php` and `tests/admin/class-w3tc-wp-cron-config-test.php`. Those files were untracked in the local working tree, so nothing was committed or pushed.

The underlying point was valid though: those paths were untracked *and* unignored, so they were one `git add -A` away from landing in the repo. `1e1c9dc6` adds targeted `.gitignore` entries for the generated agent scratch directories. The entries are deliberately narrow rather than a blanket `.claude/` or `tmp/` ignore, since `.claude/skills/` and `tmp/sec-ideas/_profile.md` are legitimately tracked and must stay that way.

Verified after the change: all four paths resolve through `git check-ignore`, all 9 previously tracked files under those trees remain tracked, and `git status` reports no remaining untracked, unignored paths.

cssjoe (approved):
Reviewed and tested on `9817a2b`. Approve.

The five Purge via WP-Cron groups are now in the config schema, so Enable / Start Time / Interval survive save on General and the per-cache pages. PHPUnit `W3tc_Wp_Cron_Config_Test` is green (6 tests, 63 assertions). Local save-path checks also covered Daily, Twice Daily, disable, and the page/db/minify/object cache pages.

The Bugbot note about agent scratch files does not apply to this SHA (only `ConfigKeys.php` and the new test file).

Optional follow-up, not blocking: declare `enum` on the interval keys to match the select-box values (`hourly`, `twicedaily`, `daily`, `weekly`).

PR #1397, opened 2026-09-16
ENG3-1072: Skip unchanged w3tc_state option writes

cssjoe:
https://imh-internal.atlassian.net/browse/ENG3-1072

https://wordpress.org/support/topic/w3tc_state-writing-to-db-on-every-admin-click-multisite/

## Summary
- Skip `update_option` / `update_site_option` in `ConfigState::save()` when the encoded `w3tc_state` payload has not changed.
- Keeps real state changes writing as before. Does not move this data into object cache.

## Test plan
- [ ] `php tests/test-config-state-save.php`
- [ ] In network admin, confirm a no-op save (for example rewriting an already-set notice flag) no longer produces a `w3tc_state` Options API write
- [ ] Confirm a real state change (dismiss a notice, save settings that update state) still persists


ConfigState.php
```diff
@@ -213,17 +224,24 @@ public function save() {
...
+			\update_site_option( 'w3tc_state', $encoded );
 		} else {
 			if ( ! \function_exists( 'update_option' ) ) {
 				return;
 			}
-			\update_option( 'w3tc_state', \wp_json_encode( $this->_data ) );
+			\update_option( 'w3tc_state', $encoded );
 		}
+
+		$this->_saved_encoded = $encoded;
```
jacobd91: **Medium:** `_saved_encoded` is advanced even when `update_option()` / `update_site_option()` returns `false`. A transient failed write is then treated as saved, so another `save()` on this instance is skipped; before this change, each call retried the Options API. A focused reproduction returns only one write attempt across two `save()` calls when the stubbed API fails. Please update the baseline only after a successful write and add a failure-then-retry case for both option paths.

jacobd91 (changes requested):
Acceptance summary:
- Standalone ConfigState suite: 11 passed
- W3TC admin smoke: 22 passed
- Authenticated network-admin smoke: 2 passed
- Direct blog and multisite Options API checks: unchanged saves made no additional writes; changed values persisted and were restored
- General multisite profile: network admin, primary admin, and front end passed; the REST index returned an environment-level 404 outside this change

Bugbot found no issues, and the security pass found no medium-or-higher issues. One reliability regression needs correction before approval.

ConfigState.php
cssjoe: Fixed in 395e8f1f.

`save()` now captures the return value of `update_option()` / `update_site_option()` and only advances `_saved_encoded` when the write reports success, so a failed write is retried on the next `save()` exactly as it was before this change.

`tests/test-config-state-save.php` gained failure-then-retry coverage for both option paths. The stubs take a `cssave_blog_fail` / `cssave_site_fail` toggle that still counts the attempt but returns `false` without storing, so each path asserts: one attempt while failing, a second attempt on the next `save()`, a third attempt that persists the value once the Options API recovers, and no further attempts after that success.

Suite is 19 passed, 0 failed (`php tests/test-config-state-save.php`).

jacobd91: Confirmed on rereview of `395e8f1f`. Failed writes no longer advance `_saved_encoded`; blog and site stubs retry until success, then stop. Standalone suite is now 19/19.

jacobd91 (approved):
Rereview of `395e8f1f` closes the failed-write retry finding.

- `_saved_encoded` now advances only after a successful Options API write
- Failure-then-retry coverage exists for both blog and site options
- Standalone suite: 19 passed
- W3TC admin smoke: 22 passed
- Authenticated network-admin smoke: 2 passed
- Direct blog and Multisite Options API checks: unchanged saves made no additional writes; changed values persisted and were restored
- Local Bugbot: no findings; security review: no medium+ issues

Approving.

PR #1398, opened 2026-09-17
ENG3-1110: Preserve custom attachment paths during CDN push

jacobd91:
https://imh-internal.atlassian.net/browse/ENG3-1110

## Summary
- preserve nested attachment directories when building CDN upload descriptors
- reject traversal and paths outside the WordPress uploads directory
- cover flat, year/month, custom nested, metadata thumbnail, absolute, and Windows-style paths

## Test plan
- [x] Focused WordPress PHPUnit suite (11 tests, 56 assertions)
- [x] PHP syntax checks for changed PHP files
- [x] Full branch diff whitespace check
- [x] Human review and CDN acceptance test

Fixes #69

---

> [!NOTE]
> **Medium Risk**
> Changes how local/remote CDN paths are derived for all attachment pushes; incorrect rejection could skip uploads, while the prior basename-only logic could mis-upload nested media.
> 
> **Overview**
> **CDN attachment path normalization** no longer collapses paths to a year/month basename. Relative paths keep their full structure (including custom folders like `prophoto/galleries/...`), and absolute paths under the uploads `basedir` are stripped to a safe relative path.
> 
> **Security hardening** rejects empty or invalid input, null bytes, `..` segments, and paths outside uploads (including prefix tricks and Windows drive-letter casing on containment checks). `get_files_for_upload` and library export skip building descriptors when normalization returns an empty string.
> 
> **Tests** add PHPUnit coverage for flat, nested, absolute, traversal, metadata thumbnails, and Windows-style paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6408052829c3e6bcdcecd379eef6a9a3da51a8ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1400, opened 2026-09-17
ENG3-1111: Cache selected external scripts locally

### LINKED ISSUE #57: [Feature] Cache external scripts
Idea submission from [WP.org Support forum](https://wordpress.org/support/topic/cache-external-ressources/)

**Description**
Allow caching of external scripts, specifically Google Analytics for this submission, by downloading a copy of the script and storing it locally.  Check for updates on a regular basis (~24 hrs) to check for modifications, and if any are found then download a new copy.  User provided [Cache External Scripts](https://wordpress.org/plugins/cache-external-scripts/) as an example for a plugin that currently has this functionality.

We could also extend this feature once implemented and allow users to specify a variety of external script sources to cache in addition to GA.

**What does this solve?**
This could improve scores from site speed testing websites that show poor performance results for 'Leverage browser caching' when external resources are gathered.

jacobd91:
https://imh-internal.atlassian.net/browse/ENG3-1111

## Summary
- cache explicitly selected external JavaScript for minification
- require HTTPS, public addresses, and same-host redirects on every hop
- bound downloads and validate responses before atomic cache replacement
- retain stale cache or original script tags when retrieval fails

## Test plan
- [x] Focused WordPress PHPUnit suite (18 tests, 51 assertions)
- [x] PHPCS and PHP syntax checks
- [x] Full branch diff whitespace check
- [ ] Human review and acceptance test with a selected external script

Fixes #57

---

> [!NOTE]
> **Medium Risk**
> Introduces server-side outbound HTTP for external JS with SSRF-style controls, but it touches the minify request path and page-output rewriting; misconfiguration or policy gaps could affect script loading or cache integrity.
> 
> **Overview**
> Adds **local caching for whitelisted external JavaScript** so auto-minify can rewrite script tags only when a usable cache entry already exists, avoiding outbound fetches during HTML generation.
> 
> **Precache / minify handler:** JavaScript downloads are gated separately from CSS—**HTTPS + public hosts only**, refresh lifetime clamped between one hour and one week, **5s timeout** and **1MB cap**, **JavaScript `Content-Type` validation**, and **atomic cache updates** via a temp file (failed refreshes keep the previous file). New read-only `get_cached_external_script()` backs `is_external_script_cached()`. Minify ID generation uses `remote_source_type()` so query-string URLs like `app.js?ver=1` still resolve as JS/CSS.
> 
> **HTTP layer:** `Util_Http::download()` gains optional policies (**public-only**, **HTTPS-only**, **same-host redirects**, size limit, response validator) applied on every redirect hop; redirect resolution now uses `WP_Http::make_absolute_url`.
> 
> **Auto JS:** Whitelisted external `src` URLs are minified only on cache hit; otherwise the original tag is left intact.
> 
> PHPUnit coverage expands for precache, redirects, stale cache, and auto-minify behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2b77e96109c7daf44ebeed66760f1d5874e0268. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Util_Http.php
```diff
@@ -102,14 +102,44 @@ public static function download( $w3tc_url, $w3tc_file, $args = array() ) {
...
-		 * with {@see Util_Url::is_allowed_outbound_url()} before the next
-		 * request. Automatic redirect following would skip that check.
+		 * against the requested policy before the next request.
 		 */
 		$max_redirects = 5;
 		for ( $hop = 0; $hop <= $max_redirects; $hop++ ) {
-			if ( ! Util_Url::is_allowed_outbound_url( $w3tc_url ) ) {
+			if ( $https_only && 'https' !== \wp_parse_url( $w3tc_url, PHP_URL_SCHEME ) ) {
+				return false;
+			}
```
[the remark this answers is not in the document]
jacobd91: Fixed in 6d6de70b and 238759b1. Protocol-relative redirects now inherit the scheme of the current request hop, while RFC-compatible resolution handles query-only, parent-relative, root-relative, and path-relative locations. Follow-up coverage also protects versioned remote JS/CSS URLs. Focused tests pass on PHP 7.4 and PHP 8.5.

Minify_AutoJs.php
```diff
@@ -250,6 +250,9 @@ private function process_script_tag( $script_tag, $script_tag_number ) {
 		$step1_result = $this->minify_helpers->is_file_for_minification( $script_src, $w3tc_file );
 		if ( 'url' === $step1_result ) {
 			$w3tc_file = $script_src;
+			if ( ! $this->minify_helpers->precache_external_script( $script_src ) ) {
+				$step1_result = '';
+			}
 		}
```
cssjoe, line 256: Auto Minify HTML generation now waits on `precache_external_script()` before rewriting the tag. On a cache miss or refresh that can stall page output for up to `EXTERNAL_JS_TIMEOUT` per selected script.

Previously this rewrite did not fetch during HTML generation; the minify request path already retrieves the remote file.

Please leave the original `<script src>` in place until a usable cached copy already exists, and refresh from the minify handler (or another non-HTML path). Add coverage that a miss keeps the original tag without an HTML-path download wait, and that a cache hit rewrites without fetching again.

Minify_MinifiedFileRequestHandler.php
```diff
@@ -858,31 +866,86 @@ public function debug_error( $error ) {
...
+		$body = isset( $response['body'] ) && \is_string( $response['body'] ) ? $response['body'] : '';
+		if ( '' === $body ) {
+			return false;
+		}
+
+		$content_type = \strtolower( \trim( \wp_remote_retrieve_header( $response, 'content-type' ) ) );
+
+		return 0 !== \strpos( $content_type, 'text/html' ) &&
+			0 !== \strpos( $content_type, 'application/xhtml+xml' );
+	}
```
cssjoe, line 1053: `validate_external_javascript_response()` currently rejects only HTML-ish `Content-Type` prefixes. A 200 with an empty type, `application/json`, or `text/plain` is still stored and can land in the minify output.

Please accept only JavaScript content types (`application/javascript`, `text/javascript`, and the usual aliases). Extend `invalid_javascript_response_provider` with missing `Content-Type` and `application/json` so that contract stays covered.

cssjoe (changes requested):
Not merge-ready on 238759b — two medium items (inline). Focused PHPUnit passed locally (26 tests, 79 assertions); syntax and PHPCS on the product files were clean. Please address the notes and rereview.

Minify_AutoJs.php
jacobd91: Fixed in a2b77e96. Auto Minify now performs a cache-only lookup during HTML generation: misses preserve the original script tag, hits rewrite from the existing atomic cache, and retrieval/refresh remains in the minify request handler. Added miss/hit tests that assert zero HTTP requests.

Minify_MinifiedFileRequestHandler.php
jacobd91: Fixed in a2b77e96. Response validation now accepts only JavaScript MIME types (including common aliases and optional parameters). Missing Content-Type and application/json are covered as rejected responses, while stale valid cache remains intact.

jacobd91:
Addressed both requested changes in a2b77e96:

- HTML generation now performs a cache-only lookup; a miss keeps the original script tag and a hit rewrites without network I/O.
- External responses now require an allowlisted JavaScript Content-Type; missing and JSON types are rejected.

Validation: focused WordPress PHPUnit 30 tests / 88 assertions; PHPCS clean on all four changed files; PHP syntax and diff checks clean.

jacobd91:
CI follow-up on a2b77e96: the branch build passed, as did conftest, PHPCS, and Semgrep. The Travis pull-request build failed only on GitHub’s synthetic merge commit; current master is also red in Travis since #1398, and Travis restricts the failing job logs. This is not reproducible on the feature branch and is tracked here as an unrelated base-branch failure pending master recovery.

Minify_AutoJs.php
```diff
@@ -250,6 +250,9 @@ private function process_script_tag( $script_tag, $script_tag_number ) {
 		$step1_result = $this->minify_helpers->is_file_for_minification( $script_src, $w3tc_file );
 		if ( 'url' === $step1_result ) {
 			$w3tc_file = $script_src;
+			if ( ! $this->minify_helpers->is_external_script_cached( $script_src ) ) {
```
cssjoe, line 253: HTML generation is correctly cache-only now, but a miss still drops this URL from the minify file list, so the minify request handler never receives it and never calls `_precache_file()`. After first use or a minify cache flush, selected scripts stay on their original tags.

Please warm the cache off the HTML path (shutdown, cron, or settings save) without blocking page output, keep the original tag until a hit, and add coverage that a miss later produces a usable cache file without an HTML-path download wait.

cssjoe (changes requested):
Not merge-ready on a2b77e96 — one high item (inline). Prior notes on HTML-path retrieval and JavaScript Content-Type allowlisting are addressed. Focused PHPUnit passed locally (30 tests, 88 assertions); syntax and PHPCS on the product files were clean. The Travis pull-request job is red on GitHub’s merge commit because current master fails `yarn run hooks:check` (`docs/hooks.md` line numbers); the feature-branch Travis build passed. Please address the remaining note and rereview.

Minify_AutoJs.php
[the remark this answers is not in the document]
jacobd91: Good catch — fixed in 843972ec. The cache-only check alone did leave the cache permanently cold, since a miss drops the URL from the minify file list and the minify handler never sees it.

A miss now queues retrieval on a WP-Cron single event (`w3tc_minify_precache_external_script`) and still leaves the original script tag in place, so page output never waits on the network. The cron callback re-checks the admin allowlist before fetching, and reuses the same HTTPS/public-host/size/MIME policy. Once a copy exists the tag is rewritten and the minify request handler keeps it refreshed, so this only covers the first fetch and recovery after a cache flush.

Coverage added for queue-on-miss, no-queue when fresh, no-queue for ineligible URLs, and an end-to-end warm where the queued fetch populates the cache and the next lookup rewrites with zero HTTP calls. Verified the wiring test fails without the fix. Focused PHPUnit: 35 tests / 106 assertions; PHPCS clean.

jacobd91:
Pushed 843972ec to fix a regression Bugbot caught in my previous commit.

Making the HTML path a cache-only check satisfied the "no network during page generation" requirement but broke the feature: a cache miss drops the URL from the minify file list, so the minify handler never receives it and nothing ever warms the cache. Selected scripts would have stayed on their original tags permanently, including after each minify cache flush.

A miss now queues a WP-Cron single event to fetch the first copy, keeping page output free of network I/O. Refresh of an already-cached script still happens on the minify request path as requested. The cron callback re-validates the URL against the admin allowlist before fetching.

Focused PHPUnit: 35 tests / 106 assertions. PHPCS, syntax, and whitespace checks clean.

Minify_AutoJs.php
jacobd91: Agreed, and fixed in 843972ec — you and Bugbot landed on the same gap.

Warming now happens off the HTML path via a WP-Cron single event (`w3tc_minify_precache_external_script`). A miss queues that event and leaves the original `<script src>` untouched, so page output still never waits on the network. The cron callback re-checks the admin allowlist before fetching and reuses the same HTTPS/public-host/redirect/size/MIME policy. Refresh of an already-cached script continues to happen on the minify request path, so cron only covers the first fetch and recovery after a minify cache flush.

Coverage added:
- a miss queues retrieval and makes no HTML-path request
- the queued retrieval later produces a usable cache file, and the next lookup rewrites with zero HTTP calls
- no queueing when the cache is fresh, or for ineligible URLs

I also confirmed the wiring test fails without the fix. Focused PHPUnit: 35 tests / 106 assertions; PHPCS and syntax clean.

Thanks for pinning down the Travis failure as master `hooks:check` — that matches what I saw, so it stays unrelated to this branch.

cssjoe (approved):
Merge-ready on 843972e — prior high (first copy never retrieved once HTML generation became cache-only) is addressed. A miss now queues WP-Cron retrieval, leaves the original script tag, and re-checks the admin selection before fetching; refresh of an existing copy stays on the minify request path. Focused PHPUnit passed locally (35 tests, 106 assertions); syntax and PHPCS on the product files were clean. The Travis pull-request job is still red on GitHub’s merge commit because current master fails `yarn run hooks:check`; the feature-branch Travis build passed.

cssjoe:
Merge-ready on 843972e. Prior high is addressed: a miss queues WP-Cron retrieval, keeps the original script tag, and re-checks the admin selection before fetching; refresh of an existing copy stays on the minify request path.

Focused PHPUnit passed locally (35 tests, 106 assertions); syntax and PHPCS on the product files were clean. Approved. The Travis pull-request job remains red on GitHub’s merge commit because current master fails `yarn run hooks:check`; the feature-branch Travis build passed.

